PR #13, opened 2021-07-17
add vector/scalar math operations; add sum,avg aggregate functions

mikeizbicki:
Thanks for the excellent library!

I had need for scalar math functions and aggregate functions while using your code, so I've added them into the repo (along with appropriate test cases).

The main thing I'm not sure about is how you want the code in `sql/vector.sql` formatted.  I've added all my additions at the bottom since that was easiest for me, but I'm not sure if you want it spread out throughout the code in different spots.  Also, I had to manually add code to both the `sql/vector.sql` file and `sql/vector--0.1.7--0.1.8.sql` files in order for it to work... this seems super error prone... is there not a way to automatically rebuild the `sql/vector.sql` automatically from all the version bumping files?


ankane:
Hey @mikeizbicki, thanks for the PR, including tests! Can you explain more about the use case for the functions?

(also, for automatically building `sql/vector.sql`, I'm not aware of a way to do this)

mikeizbicki:
My main objective was to be able to take the average of a group of vectors.  I only needed the scalar `/` (for non-aggregate queries, something like `(a+b)/2`) and `avg` for aggregate queries, but I figured I may as well implement the other scalar ops and `sum` while I was at it since it wasn't really any more work and they seemed like they'd be useful.

ankane:
What problem are you using `(a+b)/2` and `avg` for? And can you share the full queries that use them? Trying to understand the high-level use case better.

mikeizbicki:
I'm building a search engine that uses postgres as the backend.  The text search is done using standard GIN/RUM indexes, but there's also a sentiment search that is based on wordvectors.  There is a record in the db for each word/doc pair (hence the very large amount of records and why thinking about more compact storage formats in #14), but the data is partitioned on the words, so each index only stores at most about 10^6 vectors inside of it.

I am taking the average of these vectors in order to measure how average sentiment changes over time.  The actual queries are part of a rollup table framework that I've built and are machine generated.  Because they're machine generated, it's a bit difficult to extract them out and paste them here, and they're also quite long and complicated (300+ lines) and wouldn't make much sense outside of the domain context.

My goal is that by the end of the summer I'll have a prototype online, and I'd be happy to ping you when that's up so you can see more details.

ankane:
Yeah, would be great to see it when it's ready.

ankane:
Cleaning up PRs, but happy to revisit (all or parts of) this functionality if there are common use cases for it.

mikeizbicki:
I believe the functionality is ready to be merged, and I don't understand why the pr hasn't been accepted.

ankane:
The use case above seems pretty specific, so I don't think it's common enough to add right now.

mikeizbicki:
I happen to have a specific use case, but basic vector math ops seem pretty widely applicable to me.

PR #18, opened 2021-11-11
modify Makefile,add NOT USE PGXS support

wenshengzhang:
Support PostgreSQL compilation scenarios that do not use PGXS

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,9 +29,15 @@ sql/$(EXTENSION)--$(EXTVERSION).sql: sql/$(EXTENSION).sql
 
 EXTRA_CLEAN = sql/$(EXTENSION)--$(EXTVERSION).sql
 
+ifdef USE_PGXS
 PG_CONFIG ?= pg_config
 PGXS := $(shell $(PG_CONFIG) --pgxs)
 include $(PGXS)
+else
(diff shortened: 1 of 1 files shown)


ankane:
Hey @wenshengzhang, thanks for the PR! Can you explain more about the use case?

wenshengzhang:
When I installed postgresql through source code compilation, I expected to complete the compilation and deployment of pgsql core and all pgsql extensions through make world and make install-world in the source directory of postgresql. At this time, we actually cannot pass pg_config - pgxs gets the path of "pgxs.mk".
The solution is: move pgvector to the contrib directory of the Postgrersql source code and modify the Makefile so that it can find the mk file. Just like the modification in PR.

Thank you for opening pgvector, I am using it and it is great.

ankane:
Thanks for the additional context. I'm not sure there's any benefit to compiling it with Postgres rather than after, so I don't think it's a pattern I'd like to support

wenshengzhang:
> Thanks for the additional context. I'm not sure there's any benefit to compiling it with Postgres rather than after, so I don't think it's a pattern I'd like to support

Okay, I agree, and I will continue to follow pgvector . :-)

ankane:
Sounds good, thanks again for the suggestion/PR

PR #24, opened 2022-03-14
Update example to be runnable

x4m:
Hi! Thanks for the nice lib! That's really great that we have open source aNN extension.
Currently signature example does not work on my machine:
```
# CREATE TABLE table (column vector(3));
ERROR:  42601: syntax error at or near "table"
```
Here's basic attempt to fix the problem.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,9 +3,9 @@
 Open-source vector similarity search for Postgres
 
 ```sql
-CREATE TABLE table (column vector(3));
-CREATE INDEX ON table USING ivfflat (column vector_l2_ops);
-SELECT * FROM table ORDER BY column <-> '[1,2,3]' LIMIT 5;
+CREATE TABLE example_table (example_column vector(3));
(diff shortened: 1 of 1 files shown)


x4m:
Either way we can quote identifiers
```
sas-7je7f6ykc8oubeev/postgres M # CREATE TABLE "table" ("column" vector(3));
CREATE TABLE
Time: 0.749 ms
sas-7je7f6ykc8oubeev/postgres M # CREATE INDEX ON "table" USING ivfflat ("column" vector_l2_ops);
CREATE INDEX
Time: 2.317 ms
sas-7je7f6ykc8oubeev/postgres M # SELECT * FROM "table" ORDER BY "column" <-> '[1,2,3]' LIMIT 5;
 column 
--------
(0 rows)
```

ankane:
Hey @x4m, thanks for the suggestion. Just updated them.

PR #26, opened 2022-05-24
Use blis for vector operations

maparent:
This is very tentative, needs benchmarks, documentation, etc.
But it should speed up many operations.
I was curious whether this is something you want to explore.
Regards

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,8 +14,13 @@ OPTFLAGS = -march=native
 # Mac ARM doesn't support -march=native
 ifeq ($(shell uname -s), Darwin)
 	ifeq ($(shell uname -p), arm)
-		OPTFLAGS =
+		OPTFLAGS = 
 	endif
+endif
+
+ifeq ($(USE_BLIS), yes)
+	PG_CFLAGS += -DBLIS=1 $(shell pkg-config --cflags blis)
+	PG_LDFLAGS += $(shell pkg-config --libs blis)
 endif
 
 # For auto-vectorization:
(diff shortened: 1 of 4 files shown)


ankane:
Hey @maparent, thanks for the PR! It's a neat idea. Would be good to see how performance compares for each of the distance calculations to understand if it's worth the extra complexity (the functions are currently vectorized with auto-vectorization, but not sure how it compares to BLIS).

maparent:
First: delighted to hear you find this worth pursuing!
Second: What are your acceptance conditions?
Offhand, I think we'd want

1. Benchmarks
2. Git action to test the blis option
3. Some documentation

Do you see anything else?

ankane:
That looks reasonable, but need to see if the performance justifies the complexity first.

maparent:
One last comment: I also compared OpenMP, with even worse results on my (admittedly simplistic) test.

mikeizbicki:
The real advantage of BLAS libraries is when you get to the level 3 (matrix-matrix) operations.  Matrix multiplies need a lot of optimizations to be fast (selecting the right version of strassen's algorithm, and setting appropriate block sizes for cache, etc), but vector-vector operations don't really need anything special because it's just a for loop.  I could see some uses for matrix inside postgres, in which case BLAS would be a huge win, but with just vectors it's not at all surprising to me that native optimizing compilers generate faster code.

ankane:
Also, if it speeds up vector-matrix operations, there's a chance it could be useful for indexes, but probably won't be as straightforward as the changes above.

PR #49, opened 2022-12-08
Working Windows build

mmcauliffe:
Here are the changes necessary for getting pgvector to build on Windows.  Using the more recent commits has removed the need to patch for /tmp and random(), so that should all be working.

The only major thing that I'm unsure of as a non-C programmer is changing any `FLEXIBLE_ARRAY_MEMBER` to `1`, it seems to work, but I'm not sure the best practices.  From https://stackoverflow.com/questions/44689293/c-code-using-double-struct-hack-flexible-array-member-in-visual-studio, it seems like `float x[];` should be fine, but it was raising errors until I changed it to `float x[1];`, but maybe there's a better way of doing it?


ankane:
Hi @mmcauliffe, thanks for the PR and driving this!

Added experimental support with `nmake` in the commit above (to avoid a dependency on CMake and supporting multiple build systems on Linux / Mac).

```sh
nmake /F Makefile.win
nmake /F Makefile.win install
```

Please let me know if it works for you and the conda-forge case. You can use `PGROOT` to specify the path to the Postgres installation (ideally it'd use `pg_config`, but need to spend some time getting shell commands working with `nmake`).

A few other notes:

- 9c9489e88839594ad95519442845bf50e7b96e75 fixes the nested `FLEXIBLE_ARRAY_MEMBER` issue
- [Auto-vectorization](https://learn.microsoft.com/en-us/cpp/parallel/auto-parallelization-and-auto-vectorization?#auto-vectorizer) is supported for performance

mmcauliffe:
Awesome, looks like it's building correctly, barring a couple of issues.

I do have an error in `test input`:

```
diff -w -U3 C:/Users/michael/Documents/Dev/pgvector/test/expected/input.out C:/Users/michael/Documents/Dev/pgvector/results/input.out
--- C:/Users/michael/Documents/Dev/pgvector/test/expected/input.out	2022-12-09 15:23:33.514546500 -0800
+++ C:/Users/michael/Documents/Dev/pgvector/results/input.out	2022-12-09 15:29:55.170055100 -0800
@@ -12,8 +12,8 @@
 
 SELECT '[1.23456]'::vector;
   vector   
------------
- [1.23456]
+----------------------
+ [1.2345600128173828]
 (1 row)
 
 SELECT '[hello,1]'::vector;
```

And then install says that everything is up to date, but doesn't copy over any files to the conda directory I specified for `PGROOT`.

I figure these might already be on your radar, but figured I'd mention them.  Once a new release with the windows build is there with nmake, I'll transition the feedstock over to it!

ankane:
Thanks for checking. I'm not sure how to reproduce either issue, unfortunately.

CI is passing for Windows ([code](https://github.com/pgvector/pgvector/blob/7314d8b05ac0dd6070419170bb92fbea3789237a/.github/workflows/build.yml#L56-L59), [logs](https://github.com/pgvector/pgvector/actions/runs/3655078767/jobs/6176080533)).

What version of Postgres is that with?

ankane:
Hey @mmcauliffe, just released 0.4.0. Let's move this to a new issue if you're still seeing issues with the Makefile.

mmcauliffe:
Awesome, looks like everything's working well on conda forge: https://github.com/conda-forge/pgvector-feedstock/pull/6.  I added arm64 build along with using the nmake builds.

ankane:
Sweet, that was super fast 🔥 Thanks @mmcauliffe!

PR #74, opened 2023-03-22
Add instruction to build docker in Apple Silicon

felix98765:


ankane:
Hi @felix98765, thanks for the PR, but the existing command works on Apple Silicon.

PR #79, opened 2023-04-01
Update readme to show example usage of PG_CONFIG

wlaurance:
I ran into this issue because I have multiple installations of postgres on my local machine (through brew and Postgres.app)


ankane:
Thanks @wlaurance!

PR #86, opened 2023-04-06
Ivf index prefetch

knizhnik:
Implement prefetch for index scan. It can significantly increase query execution speed if index doesn't fit in memory.


ankane:
Hi @knizhnik, thanks for the PR, and sorry for the delay. Will dig into this and see how the different aspects affect performance. If you have any benchmarks, please share.

knizhnik:
> Hi @knizhnik, thanks for the PR, and sorry for the delay. Will dig into this and see how the different aspects affect performance. If you have any benchmarks, please share.

Hi @ankane !
Prefetching is most critical when working set doesn't fir in memory (neither in shared buffers, neither in file system cache).
I am adopting pgvector extension for Neon (https://neon.tech/) - it is serverless Postgres with separated compute and storage.
In this case time of roundrip between compute and pageserver is very critical. In AWS if both compute and paeserver are located in the same AZ it is about 0.2msec. In this case prefetcing of pages in ivf_index allows to increase speed about 7 times -   5198msec vs. 777 msec for [GIST](http://corpus-texmex.irisa.fr/) with 1M records each with 960 dimensions from  https://github.com/erikbern/ann-benchmarks. Unfortunately it is still ~30 times slower comparing with case when all data fits in shared buffers (23msec).

ankane:
Hi @knizhnik, sorry for not following up sooner. I think the four parts of this are:

1. Changing the buffer access strategy
2. Adding multiple (sequential) pages on inserts 
3. Prefetching pages for `GetScanItems`
4. Prefetching heap pages

For 1, the current strategy was added to prevent [cache wiping](https://momjian.us/main/blogs/pgblog/2020.html#May_13_2020), but from the article, Postgres only uses `BAS_BULKREAD` when it expects to access more than 25% of the shared buffers, so we should probably do the same.

For 2, makes sense to me (related: #133).

Let's break 1 and 2 into separate PRs to discuss.

For 3 and 4, I'm not sure I'd like to add the complexity right now, but to avoid prefetching extra pages with 3, you'll want to use `IVFFLAT_LIST_EXTEND_QUANTUM` for the index build as well. You may also want to use a custom index version and prefetch conditionally to avoid prefetching extra pages with existing indexes.

knizhnik:
> For 1, the current strategy was added to prevent [cache wiping](https://momjian.us/main/blogs/pgblog/2020.html#May_13_2020), but from the article, Postgres only uses `BAS_BULKREAD` when it expects to access more than 25% of the shared buffers, so we should probably do the same.

Postgres is using  `BAS_BULKREAD` for seqscan but here we perform index scan.
May be my assumption is wrong, but all embedding datasets I saw (for example fro ann-banchmark) are not so large and  should fit in memory. In this case filling shared buffers with this ivfflat data should not cause cache wiping and lead to faster prewarming.
Right now it is not possible to prewarm infflat index without using special extension (pg_prewarm).

> 
> Let's break 1 and 2 into separate PRs to discuss.
> 
> For 3 and 4, I'm not sure I'd like to add the complexity right now, but to avoid prefetching extra pages with 3, you'll want to use `IVFFLAT_LIST_EXTEND_QUANTUM` for the index build as well. You may also want to use a custom index version and prefetch conditionally to avoid prefetching extra pages with existing indexes.

I agree with you. Such aggressive prefetching is needed mostly for Neon architecture.
In case of vanilla Postgres we can rely on OS read-ahead mechanism.

jkatz:
From my testing, I'm -1 on changing from `BAS_BULKREAD` to `BAS_NORMAL`, though I'd like to test a bit more so I can bring numbers I'm more confident in. For queries under the defaults (i.e. `ivfflat.probes` is `1`), this likely makes little difference. However, increasing the number of probes requires loading more pages into memory, both indexes and heap, and the benefits of `BAS_BULKREAD` becoming noticeable in this case as the vector dimensionality increases.

Assuming that "datasets fit into memory" may work on systems that are exclusively processing vector data, but my current experience from seeing people integrating pgvector is that it exists alongside current workloads.

knizhnik:
> From my testing, I'm -1 on changing from `BAS_BULKREAD` to `BAS_NORMAL`, though I'd like to test a bit more so I can bring numbers I'm more confident in. For queries under the defaults (i.e. `ivfflat.probes` is `1`), this likely makes little difference. However, increasing the number of probes requires loading more pages into memory, both indexes and heap, and the benefits of `BAS_BULKREAD` becoming noticeable in this case as the vector dimensionality increases.
> 
> Assuming that "datasets fit into memory" may work on systems that are exclusively processing vector data, but my current experience from seeing people integrating pgvector is that it exists alongside current workloads.

Please notice that Postgres is using ring buffer strategies only for seqscan, not for index scan.
I wonder why ivfflat index scan is considered to be more close to seqscan rather than to index scan?
IMHO Postgres use very strange approach for buffer eviction when it is throwing away pages from shared buffers even if there are free buffers. It leads to worse performance even if it is enough free memory at the system to keep the whole data set in memory. Without usin special extensions like pg_prewarm it is not possible to to reach maximal possible speed. Yes, file system cache may help i this case, but it is about 15% slow then access to the page in shared buffers.

So I think that BAS_BULKREAD should be used at least when expected relation size is more than 1// or better 1/2 os shared buffers, but IMHO it is better not to use BAS_BULKREAD at all, because we are actually performing index scan and let buffer manager eviction algorithm do it work.

ankane:
Cleaning up stale PRs, but may revisit 1 and 2 in the future.

PR #98, opened 2023-04-22
build: support compiling with postgres

jaki:
If trying to compile this extension with postgres by placing the extension dir in postgres's contrib/ directory, make fails trying to look for pgxs in system libraries.

    Makefile:42: /usr/lib64/pgsql/pgxs/src/makefiles/pgxs.mk: No such file or directory
    make: *** No rule to make target '/usr/lib64/pgsql/pgxs/src/makefiles/pgxs.mk'.  Stop.

Many extensions (e.g. pgaudit, pg_hint_plan, pg_stat_monitor) have an if-else in their Makefile to support both compiling with PG and compiling separately.  Add similar support here.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,9 +37,16 @@ sql/$(EXTENSION)--$(EXTVERSION).sql: sql/$(EXTENSION).sql
 
 EXTRA_CLEAN = sql/$(EXTENSION)--$(EXTVERSION).sql
 
+ifdef USE_PGXS
 PG_CONFIG ?= pg_config
 PGXS := $(shell $(PG_CONFIG) --pgxs)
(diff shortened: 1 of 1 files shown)


jaki:
I understand this is a duplicate of #18, but the code is slightly different to better match existing extensions.

As for the benefit of compiling it with postgres, yugabytedb has adopted several external extensions this way.  Possibly supporting this extension is tracked in yugabyte/yugabyte-db#16166.

ankane:
Hi @jaki, thanks for the PR. However, I have the same thoughts as https://github.com/pgvector/pgvector/pull/18#issuecomment-966742312.

PR #102, opened 2023-04-25
feature request: svm ?

NotBrianZach:
Hi, I don't want to actually merge this branch, but had been reading a tweet from karpathy where he was saying svm was better in many cases than k means for embedding search, I was just wondering why postgres doesn't have support for SVM (so I asked gpt4 to write a small example one in c, haven't checked the output just loose unintegrated proof of concept)

relevant discussion from karpathy here:

https://github.com/karpathy/randomfun/blob/master/knn_vs_svm.ipynb


ankane:
Hi @NotBrianZach, thanks for sharing. From the notebook, it looks like the query vector is needed to fit the model, which won't be available when building an index.

PR #110, opened 2023-05-03
docs: aws rds supports pgvector now

nodomain:


gmccreight:
Really exciting.  Makes selling the prod story real easy now.  Also, hi Kane!

PR #113, opened 2023-05-04
Link to canonical feature request for Google Cloud SQL

airhorns:
Before this change, the README links to a Google Support Tracker issue that is now closed and marked as a duplicate of a different issue for requesting pgvector support. If the community would like to make its voice heard and get pgvector support in Cloud SQL, we should pipe in on the right issue. That way, their teams can prioritize knowing full well how many folks want support!

### DIFF
--- a/README.md
+++ b/README.md
@@ -448,7 +448,7 @@ pgvector is available on [these providers](https://github.com/pgvector/pgvector/
 
 To request a new extension on other providers:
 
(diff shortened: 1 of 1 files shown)


ankane:
Hi @airhorns, thanks for the PR. Since the existing link has all the comments, think it'd be better to ask Google to mark the newer one as a duplicate (instead of the older one).

airhorns:
Rrrroger, left a comment over there

PR #117, opened 2023-05-08
ppc64*: correct makefile optflags

tucnak:
Currently pgvector won't build on `ppc64` or `ppc64el` systems; this hotfix follows the GCC auto-vectorization guide.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,6 +18,14 @@ ifeq ($(shell uname -s), Darwin)
 		OPTFLAGS =
 	endif
 endif
+# PowerPC
+ifneq ($(filter ppc64%,$(shell uname -m)),)
+	OPTFLAGS = -mcpu=native -mtune=native -maltivec
+	# POWER9
+	ifneq (,$(findstring POWER9,$(shell grep "POWER9" /proc/cpuinfo)))
+		OPTFLAGS += -mpower9-vector
+	endif
+endif
 
 # For auto-vectorization:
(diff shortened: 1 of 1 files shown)


ankane:
Hi @tucnak, thanks for the PR! The code looks good, but I don't have a great way to test. Can you confirm the regression and TAP tests all pass on PowerPC?

```sh
make installcheck
make prove_installcheck
```

(may need to compile Postgres with `./configure --enable-tap-tests`)

ankane:
Was able to test with Docker. fwiw, auto-vectorization seemed to work without additional flags.

```text
src/vector.c:469:2: optimized: loop vectorized using 16 byte vectors
src/vector.c:496:2: optimized: loop vectorized using 16 byte vectors
src/vector.c:521:2: optimized: loop vectorized using 16 byte vectors
src/vector.c:543:2: optimized: loop vectorized using 16 byte vectors
src/vector.c:567:2: optimized: loop vectorized using 16 byte vectors
src/vector.c:594:2: optimized: loop vectorized using 16 byte vectors
src/vector.c:630:2: optimized: loop vectorized using 16 byte vectors
src/vector.c:656:2: optimized: loop vectorized using 16 byte vectors
src/vector.c:656:2: optimized:  loop versioned for vectorization because of possible aliasing
src/vector.c:682:2: optimized: loop vectorized using 16 byte vectors
src/vector.c:682:2: optimized:  loop versioned for vectorization because of possible aliasing
src/vector.c:834:3: optimized: loop vectorized using 16 byte vectors
src/vector.c:834:3: optimized:  loop versioned for vectorization because of possible aliasing
```

Thanks again for the PR!

PR #127, opened 2023-05-12
Check bounds unconditionally in array_to_vector().

bossartn:
Presently, array_to_vector()'s call to CheckDim() is skipped when typmod != -1, which allows for bypassing VECTOR_MAX_DIM.  To fix, call Check[Expected]Dim() unconditionally.  CheckExpectedDim() takes no action when typmod == -1, so there's no need to guard it with an 'if' statement.


ankane:
Thanks @bossartn!

PR #129, opened 2023-05-17
docs: pgvector now available on Azure

mulander:
Available on:
- Azure Database for PostgreSQL Flexible Server
- Azure Cosmos DB for PostgreSQL


ankane:
Awesome, thanks @mulander!

GPF199541:
I found that pgvector with citus will reduce the recall rate；

jcrupi:
when will Azure Database for PostgreSQL Flexible Server support pgvector .5+

We are pretty stuck because we want HNSW

nachoalonsoportillo:
> when will Azure Database for PostgreSQL Flexible Server support pgvector .5+
> 
> We are pretty stuck because we want HNSW

Since mid-November version 0.5.1 is supported.

PR #133, opened 2023-05-20
Update ivfflat cost estimation to choose index for more searches

jkatz:
This makes several modification to the ivfflat cost estimation to help the query planner choose it for searches in appropriate situations.

First, the update costing algorithm estimates the number of pages to visit by estimating the number of tuples based on the ratio of lists to visit. Additionally, the updated costs sets the selectivity to the probe/list ratio, given we are only visiting the subset of tuples when the probe number is less than the number of lists.

With these updates, we can infer the startup cost from the generic cost estimator, as we've tuned the selectivity and page fetching values.

Testing on data sets of different size show that the query planner will choose the ivfflat index in additional appropriate situations instead of defaulting back to a sequential scan.

### DIFF
--- a/src/ivfflat.c
+++ b/src/ivfflat.c
@@ -68,7 +68,7 @@ ivfflatcostestimate(PlannerInfo *root, IndexPath *path, double loop_count,
 {
 	GenericCosts costs;
 	int			lists;
-	double		ratio;
+	Selectivity ratio;
 	Relation	indexRel;
 #if PG_VERSION_NUM < 120000
 	List	   *qinfos;
@@ -87,6 +87,22 @@ ivfflatcostestimate(PlannerInfo *root, IndexPath *path, double loop_count,
 
 	MemSet(&costs, 0, sizeof(costs));
 
+	indexRel = index_open(path->indexinfo->indexoid, NoLock);
+	lists = IvfflatGetLists(indexRel);
+	index_close(indexRel, NoLock);
+
+	/* Get the ratio of lists that we need to visit */
+	ratio = ((Selectivity) ivfflat_probes) / lists;
+	if (ratio > 1.0)
+		ratio = 1.0;
+
+	/*
+	 * This gives us the subset of tuples to visit. This value is passed into
+	 * the generic cost estimator to determine the number of pages to visit
+	 * during the inded scan.
+	 */
+	costs.numIndexTuples = path->indexinfo->rel->tuples * ratio;
+
 #if PG_VERSION_NUM >= 120000
 	genericcostestimate(root, path, loop_count, &costs);
 #else
@@ -94,18 +110,14 @@ ivfflatcostestimate(PlannerInfo *root, IndexPath *path, double loop_count,
 	genericcostestimate(root, path, loop_count, qinfos, &costs);
 #endif
 
-	indexRel = index_open(path->indexinfo->indexoid, NoLock);
-	lists = IvfflatGetLists(indexRel);
-	index_close(indexRel, NoLock);
+	/*
+	 * if the list selectivity is lower than what is returned from the generic
+	 * cost estimator, use that.
+	 */
+	if (ratio < costs.indexSelectivity)
+		costs.indexSelectivity = ratio;
 
-	ratio = ((double) ivfflat_probes) / lists;
-	if (ratio > 1)
-		ratio = 1;
-
-	costs.indexTotalCost *= ratio;
-
-	/* Startup cost and total cost are same */
-	*indexStartupCost = costs.indexTotalCost;
+	*indexStartupCost = costs.indexStartupCost;
 	*indexTotalCost = costs.indexTotalCost;
 	*indexSelectivity = costs.indexSelectivity;
 	*indexCorrelation = costs.indexCorrelation;
--- a/src/ivfflat.c
+++ b/src/ivfflat.c
@@ -112,9 +112,17 @@ ivfflatcostestimate(PlannerInfo *root, IndexPath *path, double loop_count,
 	genericcostestimate(root, path, loop_count, qinfos, &costs);
 #endif
 
-	/* Reduce cost for sequential pages */
-	get_tablespace_page_costs(path->indexinfo->reltablespace, NULL, &spc_seq_page_cost);
-	costs.indexTotalCost -= 0.9 * costs.numIndexPages * (costs.spc_random_page_cost - spc_seq_page_cost);
+	/* Adjust cost if needed since TOAST not included in seq scan cost */
+	if (costs.numIndexPages > path->indexinfo->rel->pages && ratio < 0.5)
+	{
+		get_tablespace_page_costs(path->indexinfo->reltablespace, NULL, &spc_seq_page_cost);
+
+		/* Change page cost from random to sequential */
+		costs.indexTotalCost -= costs.numIndexPages * (costs.spc_random_page_cost - spc_seq_page_cost);
+
+		/* Remove cost of extra pages */
+		costs.indexTotalCost -= (costs.numIndexPages - path->indexinfo->rel->pages) * spc_seq_page_cost;
+	}
 
 	/*
 	 * If the list selectivity is lower than what is returned from the generic
--- a/src/ivfflat.c
+++ b/src/ivfflat.c
@@ -72,6 +72,7 @@ ivfflatcostestimate(PlannerInfo *root, IndexPath *path, double loop_count,
 	double		ratio;
 	double		spc_seq_page_cost;
 	Relation	indexRel;
+	Cost		startupCost;
 #if PG_VERSION_NUM < 120000
 	List	   *qinfos;
 #endif
@@ -112,16 +113,18 @@ ivfflatcostestimate(PlannerInfo *root, IndexPath *path, double loop_count,
 	genericcostestimate(root, path, loop_count, qinfos, &costs);
 #endif
 
+	startupCost = costs.indexTotalCost;
+
 	/* Adjust cost if needed since TOAST not included in seq scan cost */
 	if (costs.numIndexPages > path->indexinfo->rel->pages && ratio < 0.5)
 	{
 		get_tablespace_page_costs(path->indexinfo->reltablespace, NULL, &spc_seq_page_cost);
 
 		/* Change page cost from random to sequential */
-		costs.indexTotalCost -= costs.numIndexPages * (costs.spc_random_page_cost - spc_seq_page_cost);
+		startupCost -= costs.numIndexPages * (costs.spc_random_page_cost - spc_seq_page_cost);
 
 		/* Remove cost of extra pages */
-		costs.indexTotalCost -= (costs.numIndexPages - path->indexinfo->rel->pages) * spc_seq_page_cost;
+		startupCost -= (costs.numIndexPages - path->indexinfo->rel->pages) * spc_seq_page_cost;
 	}
 
 	/*
@@ -131,7 +134,7 @@ ivfflatcostestimate(PlannerInfo *root, IndexPath *path, double loop_count,
 	if (ratio < costs.indexSelectivity)
 		costs.indexSelectivity = ratio;
 
-	*indexStartupCost = costs.indexTotalCost;
+	*indexStartupCost = startupCost;
 	*indexTotalCost = costs.indexTotalCost;
 	*indexSelectivity = costs.indexSelectivity;
 	*indexCorrelation = costs.indexCorrelation;
--- a/src/ivfflat.c
+++ b/src/ivfflat.c
@@ -7,6 +7,7 @@
 #include "ivfflat.h"
 #include "utils/guc.h"
 #include "utils/selfuncs.h"
+#include "utils/spccache.h"
 
 #if PG_VERSION_NUM >= 120000
 #include "commands/progress.h"
@@ -68,7 +69,8 @@ ivfflatcostestimate(PlannerInfo *root, IndexPath *path, double loop_count,
 {
 	GenericCosts costs;
 	int			lists;
-	Selectivity ratio;
+	double		ratio;
+	double		spc_seq_page_cost;
 	Relation	indexRel;
 #if PG_VERSION_NUM < 120000
 	List	   *qinfos;
@@ -92,14 +94,14 @@ ivfflatcostestimate(PlannerInfo *root, IndexPath *path, double loop_count,
 	index_close(indexRel, NoLock);
 
 	/* Get the ratio of lists that we need to visit */
-	ratio = ((Selectivity) ivfflat_probes) / lists;
+	ratio = ((double) ivfflat_probes) / lists;
 	if (ratio > 1.0)
 		ratio = 1.0;
 
 	/*
 	 * This gives us the subset of tuples to visit. This value is passed into
 	 * the generic cost estimator to determine the number of pages to visit
-	 * during the inded scan.
+	 * during the index scan.
 	 */
 	costs.numIndexTuples = path->indexinfo->rel->tuples * ratio;
 
@@ -110,14 +112,18 @@ ivfflatcostestimate(PlannerInfo *root, IndexPath *path, double loop_count,
 	genericcostestimate(root, path, loop_count, qinfos, &costs);
 #endif
 
+	/* Reduce cost for sequential pages */
+	get_tablespace_page_costs(path->indexinfo->reltablespace, NULL, &spc_seq_page_cost);
+	costs.indexTotalCost -= 0.9 * costs.numIndexPages * (costs.spc_random_page_cost - spc_seq_page_cost);
+
 	/*
-	 * if the list selectivity is lower than what is returned from the generic
+	 * If the list selectivity is lower than what is returned from the generic
 	 * cost estimator, use that.
 	 */
 	if (ratio < costs.indexSelectivity)
 		costs.indexSelectivity = ratio;
 
-	*indexStartupCost = costs.indexStartupCost;
+	*indexStartupCost = costs.indexTotalCost;
 	*indexTotalCost = costs.indexTotalCost;
 	*indexSelectivity = costs.indexSelectivity;
 	*indexCorrelation = costs.indexCorrelation;


ankane:
Hi @jkatz, thanks for the PR! Digging into the cost estimates, I think most of the current error comes using the random page cost (via [genericcostestimate](https://github.com/postgres/postgres/blob/REL_15_3/src/backend/utils/adt/selfuncs.c#L6566)) for all index pages when most page fetches will be sequential when the index is created.

Applying the ratio to `costs.numIndexTuples` before passing to `genericcostestimate` makes sense and will ensure `costs.numIndexPages` gets set correctly (which isn't the case now, but from [the docs](https://www.postgresql.org/docs/current/index-cost-estimation.html), it's only used for parallel scans).

I also think the selectivity change is good based on the docs (but don't fully understand how the planner is using it at this point).

Start-up cost is "the part of the total scan cost that must be expended before we can begin to fetch the first row.", which will be high for an IVFFlat index, which must scan and sort all the data before any rows are returned, so I don't think it should use the generic estimate of 0.

I've built upon the PR in the [costs branch](https://github.com/pgvector/pgvector/compare/costs) to replace some of the random page costs with the sequential cost (the branch currently assumes 90% sequential fetches). However, this will underestimate sequential fetches when the index is first created and overestimate them after a certain point as the index is updated. We could probably ensure a high ratio of sequential fetches by extending lists by multiple pages at a time (one part of #86).

Let me know your thoughts.

jkatz:
Hi @ankane,

> I think most of the current error comes using the random page cost 

That's exactly it. I should have noted that better in the PR. Before the PR, you can see this quite directly when modifying `random_page_cost`. While working through this, I've had discussions with pgvector users to lower `random_page_cost` while we work on optimizing the defaults :smile:

> I also think the selectivity change is good based on the docs (but don't fully understand how the planner is using it at this point).

Far be it from me to have any expertise on the planner, but AIUI selectivity is ultimately telling the planner how many tuples will be considered / eliminate by any qualifiers, which is typically predicates in the `WHERE` clause. However, ivfflat is a bit different, because the selectivity is coming from the number of probes, hence the recommendation for this change. Because we've lowered the selectivity, the planner knows that we don't have to consider the full set of rows in the index and likely gets some costing / optimization based on that.

The one thing I haven't figured out is how to bubble up the fact that we're only going to consider `probes / lists` rows such that it shows in the `EXPLAIN` plan.  I couldn't figure that out from the code itself, and I'm unsure if that is actually exposed from the index access method, so that may take some inquiring.

> Start-up cost is "the part of the total scan cost that must be expended before we can begin to fetch the first row.", which will be high for an IVFFlat index, which must scan and sort all the data before any rows are returned, so I don't think it should use the generic estimate of 0.

 I think logically that makes sense, but I do wonder if there's a "in theory" / "in practice" element we should consider. It makes me wonder if things are a bit flipped in a way -- AIUI MOST OF ivfflat is startup cost, which makes the "total cost" pretty cheap? Maybe it's good enough to capture the full cost in total cost?

The other behavior I've noticed is that, particularly for ivfflat indexes on higher-dimensionality vectors (e.g. 1536 dims) + lots of rows, the initial penalty of moving the pages into memory is high. In fact, it usually takes a few query runs to get them fully cached (which is something else I'd love to dive into), but once it's set it'll greatly outperform the sequential scan.

> I've built upon the PR in the [costs branch](https://github.com/pgvector/pgvector/compare/costs) to replace some of the random page costs with the sequential cost (the branch currently assumes 90% sequential fetches). However, this will underestimate sequential fetches when the index is first created and overestimate them after a certain point as the index is updated. We could probably ensure a high ratio of sequential fetches by extending lists by multiple pages at a time (one part of https://github.com/pgvector/pgvector/pull/86).

Cool -- I can test this out later today. I have a few data sets / experiments lying around to see how this affects behavior.

Thanks!

Jonathan

ankane:
The problem with setting the start-up cost low is it'll always use the index, even when a seq scan is faster (for instance, when probes = lists with multiple parallel workers).

For caching, the index scan code currently uses [BAS_BULKREAD](https://github.com/pgvector/pgvector/blob/e9c88d6f25499f17ac183dcd05af94b0f74b503b/src/ivfscan.c#L127) to avoid [cache wiping](https://momjian.us/main/blogs/pgblog/2020.html#May_13_2020), but maybe this should be revisited.

jkatz:
> The problem with setting the start-up cost low is it'll always use the index, even when a seq scan is faster (for instance, when probes = lists with multiple parallel workers).

So interestingly, in my tests with the 1MM, 1536-dim data set, the index scan was faster when the caches were warm. That said, I do agree there's some startup cost.

I'm going to try to test out your patch today -- I had some thoughts on what else we could do about startup cost, but I should test vs. speaking in hypotheticals 😄 

> For caching, the index scan code currently uses [BAS_BULKREAD](https://github.com/pgvector/pgvector/blob/e9c88d6f25499f17ac183dcd05af94b0f74b503b/src/ivfscan.c#L127) to avoid [cache wiping](https://momjian.us/main/blogs/pgblog/2020.html#May_13_2020), but maybe this should be revisited.

Cool. I also haven't actually directly measured the cache impact outside of observation, but definitely interested in seeing what we can do here.

jkatz:
OK I did some testing on two different data sets.

For a 100K, 256-dim data set with `lists = 100`,  the [modified costs](https://github.com/pgvector/pgvector/compare/costs) in your branch seem to work well. The query planner (QP) keeps choosing the index up until `ivfflat.probes = 100`, where it chooses the sequential scan. At that point, the query times between the two are about even.

However, this seems to break down for a 1MM, 1536-dim data sets with `lists = 1000`. With `ivfflat.probes = 30` (which seems to be a sweet spot in performance/recall tradeoff), the QP is picking the sequential scan (cold: 60s, warm: 5s -- I have 2 parallel workers) vs. the index (cold: 15s, warm 0.6s).

That said, this is generally better. I still feel like there is some tweaking we can do, but I think the QP is now choosing the index plan in more cases where it should be doing so.

Do we want to put the sequential fetch tuner (`0.9`) into a GUC? Arguments against doing so are 1/ who knows how permanent it will be and 2/ users can still leverage `random_page_cost` to tune the total index costs on the queries (though that would require documentation).

ankane:
Hmm, I'm having trouble reproducing, even without the patch (in which case it stops using the index between 250 and 300 probes). What are your values for `seq_page_cost` and `random_page_cost`, and what costs are you seeing from `EXPLAIN ANALYZE` with `enable_seqscan = on/off`?

jkatz:
@ankane I had the default for `seq_page_cost` (`1`) and `random_page_cost` (`4`).

Other (semi-)relevant settings:

* `shared_buffers` TO `4GB`
* `work_mem` TO `16MB`
* `effective_cache_size` TO `8GB`
* 2 core systems

I used your costing branch for the tests. While the plans are the same, here are the results after the cache warmed:

`enable_seq_scan` to `on`:

```
 Limit  (cost=19499.05..19499.63 rows=5 width=12) (actual time=5217.261..5298.944 rows=5 loops=1)
   ->  Gather Merge  (cost=19499.05..116728.14 rows=833334 width=12) (actual time=5217.259..5298.941 rows=5 loops=1)
         Workers Planned: 2
         Workers Launched: 2
         ->  Sort  (cost=18499.02..19540.69 rows=416667 width=12) (actual time=5199.207..5199.208 rows=4 loops=3)
               Sort Key: <very big vector>
               Sort Method: top-N heapsort  Memory: 25kB
               Worker 0:  Sort Method: top-N heapsort  Memory: 25kB
               Worker 1:  Sort Method: top-N heapsort  Memory: 25kB
               ->  Parallel Seq Scan on vecs  (cost=0.00..11578.33 rows=416667 width=12) (actual time=0.598..5098.805 rows=333333 loops=3)24 ms

```

`enable_seq_scan` to `off`:

```
 Limit  (cost=39265.30..39265.44 rows=5 width=12) (actual time=534.236..534.278 rows=5 loops=1)
   ->  Index Scan using vecs_data_idx on vecs  (cost=39265.30..67545.30 rows=1000000 width=12) (actual time=534.234..534.275 rows=5 loops=1)
         Order By: (data <-> <very big vector>)
```

In terms of execution times when going from cold => warm, the index was still faster.

ankane:
Thanks, can now reproduce. It looks like there's a big difference in the cost with `PLAIN` vs `EXTENDED` storage (with no index).

Plain

```text
Limit  (cost=1013129.05..1013129.63 rows=5 width=12) (actual time=13409.818..13410.621 rows=5 loops=1)
```

Extended (default)

```text
Limit  (cost=19536.79..19537.38 rows=5 width=12) (actual time=16382.932..16389.773 rows=5 loops=1)
```

`EXTENDED` has a much lower cost, even though `PLAIN` is faster, and the cost for `PLAIN` seems more accurate when compared to the cost with an index. I think this may be the root of the issue (even more than the random vs sequential page cost).

jkatz:
Nice analysis. I'll experiment a bit with `PLAIN`.

jkatz:
OK, I reproduced this and I've figured out what's happening with the costs.

When calculating the cost of a sequential scan, the planner estimates what the cost of the disk operations are using `seq_page_cost` and the number of pages in the relation:

```c
disk_run_cost = spc_seq_page_cost * baserel->pages;
```

Using the above data set, here's what happens in both types:

`EXTENDED`:

```
 relpages 
----------
     6370
```

`PLAIN`:

```
 relpages 
----------
  1000000
```

This is why when storing the vectors with `EXTENDED` storage, we're coming up with a lower cost for the sequential scan.

I don't have any suggestions just yet, as I used up my evening brainpower getting to this conclusion. If we do have access to the attribute storage type (I _think_ we can get that from `indexRel`), we could apply a discount on IVF FLAT scans when we're using non-plain storage, but I'd want to think through those ramifications.

ankane:
Thanks for digging into this. It looks like the mailing list archive confirms TOAST isn't accounted for in the seq scan cost estimate ([from 2005](https://www.postgresql.org/message-id/flat/42010C24.90301%40logi-track.com)), which is also an issue for GIN indexes ("searching in GIN indices seems in some situations to un-fairly favor Sequential Scans" - [from 2011](https://www.postgresql.org/message-id/flat/4ED68EEC.9030906%40krogh.cc)).

This seems like a tough spot for access methods. Since the seq scan cost is inaccurate, we'll need to make the index scan cost inaccurate, which will affect the comparison with other index scans for metadata filtering.

One way to do this would be:

```c
if (path->indexinfo->pages > path->indexinfo->rel->pages)
    costs.indexTotalCost *= ((double) path->indexinfo->rel->pages) / path->indexinfo->pages;
```

However, this will reduce the cost too much. I think we'll want to find a way to minimize the cost inaccuracy (basically, lower the cost just enough so it's only slightly better than a seq scan if needed). I've experimented with using `costs.numIndexPages` instead of `path->indexinfo->pages` to do this, but don't have a great solution at the moment.

jkatz:
To quote from that thread[1]:

> Seems like a bit of a can of worms ...

I think we'll likely have something imperfect but hopefully good for 90% of the cases. That said, I only have some bad ideas at the moment:

1. Set storage to `PLAIN` while `sizeof(Vector(n)) < (sizeof(Page) - sizeof(PageHeader))` (likely picking a constant like `2000`. After that, set storage to `EXTENDED`. While possible, I don't think there's a simple way to do this that wouldn't either be inconvenient for users (e.g. via a hook, which would involve having a shared preload library or an explicit `LOAD` command)  or an event trigger (which I don't know if it can adjust the column storage from within the statement -- and in particular I'm unsure about this on CREATE TABLE).
2. Look up the attribute information (e.g. via `SearchSysCacheAttName` if we can't get to it from the existing memory structures, I haven't checked) to determine if the column is using non-plain storage. If it is, then we apply a costing discount based on either the ratio you proposed above or a different formula that can better correlate to what the cost of the sequential scan would be.

One thing to note with:

```c
if (path->indexinfo->pages > path->indexinfo->rel->pages)
    costs.indexTotalCost *= ((double) path->indexinfo->rel->pages) / path->indexinfo->pages;
```

This will likely hold true even if the storage is `PLAIN` on larger vectors. Let's take the 1536-dim vectors as an example, say with 1000 lists. Given the size, the table will have a 1:1 ratio of vector to pages. This is the same with the index too, BUT the index also needs to have

1. Pages for the centroids
2. The root page

So no matter what, this will hold true, even if the cost discount is negligible in the `PLAIN` case.

It's definitely an interesting problem.

[1] https://www.postgresql.org/message-id/9501.1107388153%40sss.pgh.pa.us

jkatz:
Another bad idea: recommending to set `toast_tuple_target` to the maximum value (8160), but that is a table-wide parameter, so I'm hesitant to add that as a recommendation.

jkatz:
For a sanity check -- I did add in the commits from https://github.com/pgvector/pgvector/tree/costs into this PR, pending the outcome of this discussion.

ankane:
Pushed an update to the costs branch.

```c
/* Adjust cost if needed since TOAST not included in seq scan cost */
if (costs.numIndexPages > path->indexinfo->rel->pages && ratio < 0.5)
{
	get_tablespace_page_costs(path->indexinfo->reltablespace, NULL, &spc_seq_page_cost);

	/* Change page cost from random to sequential */
	costs.indexTotalCost -= costs.numIndexPages * (costs.spc_random_page_cost - spc_seq_page_cost);

	/* Remove cost of extra pages */
	costs.indexTotalCost -= (costs.numIndexPages - path->indexinfo->rel->pages) * spc_seq_page_cost;
}
```

It no longer converts a fraction of the random page cost -> seq page cost in all cases, but want to keep it focused on the seq scan issue. Also, the 0.5 is arbitrary, but want to allow it to consider parallel seq scans when the ratio is high enough.

Re ideas: For 1 and 3, I don't think we should change anything with the actual storage (just the costs), and I think the first condition above will account for 2.

jkatz:
@ankane This seems to work pretty on my tests with larger vectors (1536-dim, 1MM rows).

On smaller vectors (256-dim, 100K vectors, 100 lists) with `ivfflat.probes` set to 30, I'm getting a sequential scan when an index scan is outperforming it 2x.

Explain plans below:

Seq scan (45ms):

```
 Limit  (cost=16498.93..16499.52 rows=5 width=12)
   ->  Gather Merge  (cost=16498.93..26221.91 rows=83334 width=12)
         Workers Planned: 2
         ->  Sort  (cost=15498.91..15603.08 rows=41667 width=12)
               Sort Key: (('[0.05209669,0.010887866,0.084184535,0.10770733,0.018193219,0.09147514,0.10640675,0.077822134,0.08219052,0.049123086,0.080244966,0.00931643,0.018448181,0.067092516,0.059309423,0.057746902,0.046737127,0.05576068,0.07668736,0.042433735,0.054925855,0.0013007156,0.0733106,0.0038234962,0.0782525,0.04089636,0.0100064445,0.021850618,0.0020812692,0.10401375,0.03756528,0.0074024973,0.048883304,0.07666758,0.009973977,0.021406705,0.07360067,0.103970535,0.0871376,0.0616823,0.09670182,0.008877082,0.07232387,0.010308288,0.053544447,0.10446892,0.0021241193,0.008818389,0.028816868,0.09096498,0.07548272,0.008167481,0.06209789,0.061723683,0.10583522,0.0086986,0.07667107,0.009077914,0.075711325,0.07878644,0.039364733,0.089823194,0.051205512,0.042980883,0.060558636,0.033549093,0.072811075,0.08319693,0.03554955,0.054803863,0.03576427,0.10671894,0.0018799234,0.055943962,0.09986665,0.016540505,0.038176227,0.014930798,0.017459298,0.008102924,0.03562015,0.08548823,0.0033591983,0.061496153,0.103136785,0.10462727,0.085208535,0.08855632,0.05070697,0.102894455,0.088402554,0.0336774,0.080244586,0.05740841,0.043059997,0.034787815,0.07905015,0.030118033,0.053557273,0.02744842,0.076181106,0.08260156,0.0038970474,0.010516841,0.005873886,0.04998968,0.069289595,0.034933615,0.02818876,0.099745475,0.054342184,0.062566675,0.061786506,0.04939065,0.0817926,0.108007565,0.10599676,0.054797634,0.015940905,0.007752264,0.016461626,0.09412956,0.09305765,0.044617858,0.017929148,0.015482806,0.056639176,0.03201664,0.031276073,0.07671124,0.103922285,0.08524696,0.09299353,0.055925813,0.0032087783,0.04063231,0.058623344,0.079262264,0.033153415,0.017935878,0.0373429,0.031717848,0.10644514,0.051871054,0.04539223,0.08494648,0.09262338,0.025480466,0.07714622,0.02934202,0.097412795,0.10313015,0.036021437,0.02293162,0.03945242,0.025450516,0.104776375,0.10199547,0.08339471,0.041306674,0.042587966,0.04123421,0.009407661,0.0011610531,0.047035817,0.094618894,0.09516559,0.082266584,0.10030259,0.08983159,0.08120484,0.040542006,0.102485135,0.09477684,0.108873345,0.055144582,0.018733703,0.05973221,0.08083418,0.03564454,0.08630911,0.09609358,0.10930955,0.07609193,0.077442765,0.017102065,0.027433405,0.019370388,0.040732965,0.04241843,0.10600057,0.02256147,0.07718586,0.09588436,0.018009275,0.027194615,0.04400973,0.057818033,0.014628477,0.019316278,0.10512651,0.048948295,0.054104432,0.06634485,0.024690209,0.00507515,0.0016621987,0.048285224,0.07658825,0.07227534,0.028680451,0.022090446,0.10703324,0.013243557,0.01687065,0.04978868,0.07970361,0.01568018,0.08362049,0.08354457,0.062233012,0.040958095,0.04988603,0.011379617,0.067094326,0.06712714,0.055351898,0.03740438,0.012659421,0.049691193,0.0014410588,0.01072224,0.08376855,0.04625791,0.087403946,0.090959266,0.06003482,0.06576075,0.03016992,0.09364644,0.03310974,0.06530917,0.029766874,0.08764208,0.02623412,0.009101469,0.06400888,0.063115016,0.0044831694,0.0051593827,0.007643593,0.032092094,0.024881,0.094710104,0.07597842,0.0148629155]'::vector(256) <-> data))
               ->  Parallel Seq Scan on vecs_small  (cost=0.00..14806.83 rows=41667 width=12)
```

Index scan (22ms):

```
 Limit  (cost=17437.00..17439.88 rows=5 width=12)
   ->  Index Scan using vecs_small_data_idx on vecs_small  (cost=17437.00..75131.00 rows=100000 width=12)
         Order By: (data <-> '<redacted>'::vector(256))
```

However, at 20 probes it's still choosing the index scan.

ankane:
Am able to repro with those settings. Will see if there's a good way to remove a little more cost before a release, but should already be a significant improvement. Thanks for reporting, discussing, testing, and working together on the fix!

ankane:
It looks like situation above was caused by the first topic (random vs sequential cost). I think combining both approaches will give the best results, so added in the commit above.

PR #148, opened 2023-06-08
fix: input function does not allow tailing spaces

### LINKED ISSUE #147: Input function does not allow tailing spaces
```
# select '[0.24    ]'::vector(1);
ERROR:  invalid input syntax for type vector: "0.24    ]"
LINE 1: select '[0.24    ]'::vector(1);
               ^
# select '[0.24]'::vector(1);
 vector 
--------
 [0.24]
(1 row)
```

yihong0618:
this fix #147 

also add test case for it

passed CI can check: https://github.com/yihong0618/pgvector/actions/runs/5209060111

### DIFF
--- a/src/vector.c
+++ b/src/vector.c
@@ -193,7 +193,7 @@ vector_in(PG_FUNCTION_ARGS)
 		CheckElement(x[dim]);
 		dim++;
 
-		while (vector_isspace((unsigned char)*stringEnd))
+		while (vector_isspace(*stringEnd))
           stringEnd++;
 
 		if (stringEnd == pt)
@@ -217,7 +217,7 @@ vector_in(PG_FUNCTION_ARGS)
 
 	stringEnd++;
 	/* only whitespace is allowed after the closing brace */
-	while (vector_isspace((unsigned char)*stringEnd))
+	while (vector_isspace(*stringEnd))
 		stringEnd++;
 
 	if (*stringEnd != '\0')
--- a/src/vector.c
+++ b/src/vector.c
@@ -171,6 +171,9 @@ vector_in(PG_FUNCTION_ARGS)
 		x[dim] = strtof(pt, &stringEnd);
 		CheckElement(x[dim]);
 		dim++;
+		// stringEnd is space is OK to support SELECT '[   1  ,       2     ]'::vector(2);
+		while (isspace((unsigned char)*stringEnd))
+          stringEnd++;
 
 		if (stringEnd == pt)
 			ereport(ERROR,
--- a/src/vector.c
+++ b/src/vector.c
@@ -88,11 +88,11 @@ CheckElement(float value)
 
 /*
  * PostgreSQL has with the array_isspace function for the character checking which is reimplemented
- * here, as it's static, source code:
+ * as vector_isspace, as it's static, source code link:
  * https://github.com/postgres/postgres/blob/378d73ef204d0dcbeab834d52478e8cb90578ab7/src/backend/utils/adt/arrayfuncs.c#L438
  */
 static inline bool
-array_isspace(char ch)
+vector_isspace(char ch)
 {
 	if (ch == ' ' ||
 		ch == '\t' ||
@@ -168,11 +168,7 @@ vector_in(PG_FUNCTION_ARGS)
 	char	   *stringEnd;
 	Vector	   *result;
 
-	/*
-	 * Note: we currently allow whitespace between, but not within,
-	 * dimension items.
-	 */
-	while (array_isspace(*str))
+	while (vector_isspace(*str))
 		str++;
 
 	if (*str != '[')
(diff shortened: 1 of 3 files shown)


jkatz (commented):
This aligns with current PostgreSQL behavior with arrays, e.g.:

```
postgres=# SELECT '{1   , 2   }'::int[];
 int4  
-------
 {1,2}
(1 row)
```

However, I'd recommend following the convention that PostgreSQL has with the `array_isspace` function for the character checking (which would have to be reimplemented, as it's static):

https://github.com/postgres/postgres/blob/378d73ef204d0dcbeab834d52478e8cb90578ab7/src/backend/utils/adt/arrayfuncs.c#L438

Here is an example of how it is used in the `array_in` code:

https://github.com/postgres/postgres/blob/378d73ef204d0dcbeab834d52478e8cb90578ab7/src/backend/utils/adt/arrayfuncs.c#L257

yihong0618:
> This aligns with current PostgreSQL behavior with arrays, e.g.:
> 
> ```
> postgres=# SELECT '{1   , 2   }'::int[];
>  int4  
> -------
>  {1,2}
> (1 row)
> ```
> 
> However, I'd recommend following the convention that PostgreSQL has with the `array_isspace` function for the character checking (which would have to be reimplemented, as it's static):
> 
> https://github.com/postgres/postgres/blob/378d73ef204d0dcbeab834d52478e8cb90578ab7/src/backend/utils/adt/arrayfuncs.c#L438
> 
> Here is an example of how it is used in the `array_in` code:
> 
> https://github.com/postgres/postgres/blob/378d73ef204d0dcbeab834d52478e8cb90578ab7/src/backend/utils/adt/arrayfuncs.c#L257

thanks will use `array_isspace`

src/vector.c
```diff
@@ -86,6 +86,24 @@ CheckElement(float value)
 				 errmsg("infinite value not allowed in vector")));
 }
 
+/*
+ * PostgreSQL has with the array_isspace function for the character checking which is reimplemented
+ * here, as it's static, source code:
+ * https://github.com/postgres/postgres/blob/378d73ef204d0dcbeab834d52478e8cb90578ab7/src/backend/utils/adt/arrayfuncs.c#L438
+ */
+static inline bool
+array_isspace(char ch)
```
ankane, line 95: Let's name this `vector_isspace`

src/vector.c
```diff
@@ -150,6 +168,13 @@ vector_in(PG_FUNCTION_ARGS)
 	char	   *stringEnd;
 	Vector	   *result;
 
+	/*
+	 * Note: we currently allow whitespace between, but not within,
+	 * dimension items.
```
ankane, line 173: I'm not sure this comment makes sense for vectors, so let's remove it

src/vector.c
```diff
@@ -171,6 +196,9 @@ vector_in(PG_FUNCTION_ARGS)
 		x[dim] = strtof(pt, &stringEnd);
 		CheckElement(x[dim]);
 		dim++;
+		// stringEnd is space is OK to support SELECT '[   1  ,       2     ]'::vector(2);
+		while (array_isspace((unsigned char)*stringEnd))
```
ankane, line 200: No need to cast here or below

test/sql/input.sql
```diff
@@ -1,5 +1,9 @@
 SELECT '[1,2,3]'::vector;
 SELECT '[-1,2,3]'::vector;
+SELECT '[1,2,3  ]'::vector;
+SELECT '[1,     2,    3  ]'::vector;
+SELECT '[1.   , 2, 3 ]'::vector(3);
+SELECT ' [1,2,3] '::vector(3);
```
ankane, line 6: I think we can combine this into a single test

ankane (commented):
Hi @yihong0618, thanks for reporting and the PR! Looks great. Added a few comments inline (accidentally submitted the first review early).

test/sql/input.sql
yihong0618: copy that

src/vector.c
yihong0618: seems we need that, this is handle `[1   , 2    ]` which has spaces between `1 and ,`

src/vector.c
yihong0618: copy that

src/vector.c
ankane: `(unsigned char)` shouldn't be needed. Also, I think it should come after the `if (stringEnd == pt)` check.

yihong0618: copy that, thanks for the info~

yihong0618:
@ankane thanks very much for the review, the comments has been fixed now, and one CI on ubuntu seems failed, 
I am not sure its flaky test will check: https://github.com/yihong0618/pgvector/actions/runs/5217584573/jobs/9417561780

ankane:
Awesome, big thanks @yihong0618!

yihong0618:
@ankane thanks for your great repo~

PR #155, opened 2023-06-09
Support minor version selection

toddtarsi:
I think we're on like 15.1 or something, and don't want to use 15.3 locally. Would something like this be alright?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,10 @@
 ARG PG_MAJOR=15
-FROM postgres:$PG_MAJOR
+ARG PG_FULL=$PG_MAJOR
+
+FROM postgres:$PG_FULL
+
 ARG PG_MAJOR
+ARG PG_FULL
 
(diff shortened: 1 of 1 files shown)


df7cb:
https://why-upgrade.depesz.com/show?from=15.1&to=15.3

toddtarsi:
@df7cb - I get that. It's been a bit and maybe this isn't even a concern anymore. The reason for us was at the time, we couldn't because AWS hadn't caught up and was still like vetting stuff for Aurora rollout. I'm all for updating, but its often important to keep parity with what is currently deployed.

toddtarsi:
@df7cb - Thank you for addressing this PR as well as contributing to an awesome piece of tech

ankane:
Hi @toddtarsi, thanks for the PR, and sorry for the delay. I'd like to keep the Dockerfile very simple for now (but users who need this can always copy and modify it for their application).

PR #156, opened 2023-06-10
Fix compiler warnings on Ubuntu

jkatz:
The compiler was complaining about a missing include due to the addition of NaN/inf checks in 482a5f8b. Adding the include silences the warnings.


ankane:
Awesome, thanks! Will also fix that test for PRs.

PR #165, opened 2023-06-18
Modify index AM definitions to follow upstream convention

jkatz:
The upstream index access methods declare them to be `extern.` While
the current definitions seem benign, this definition is needed
when an access method may be used beyond its current scope. For
example, prior to this commit, the index costing functions in
PostgreSQL would not execute the full functions due to the
`ivfflatcostestimate` not having an `extern` definition. In this case,
this would skip determining if PostgreSQL should perform a parallel
index scan.

### DIFF
--- a/src/ivfbuild.c
+++ b/src/ivfbuild.c
@@ -636,7 +636,7 @@ BuildIndex(Relation heap, Relation index, IndexInfo *indexInfo,
 /*
  * Build the index for a logged table
  */
-IndexBuildResult *
+extern IndexBuildResult *
 ivfflatbuild(Relation heap, Relation index, IndexInfo *indexInfo)
 {
 	IndexBuildResult *result;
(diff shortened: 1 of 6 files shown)


ankane:
Hey @jkatz, thanks for another PR! What's the best way to test / see a difference with the changes? (the `parallel-index-scan2` branch seems to call `ivfflatcostestimate` twice without `extern`)

jkatz:
@ankane I don't have a great answer on this one, as I did it by putting in debug statements into PostgreSQL code. But it was before/after this set of lines:

https://github.com/postgres/postgres/blob/797f98036400bc70d4b331c5b1760a05aab59cfa/src/backend/optimizer/path/costsize.c#L589-L593

jkatz:
@ankane Alternatively, you can step through `gdb` or `lldb` to see the execution skip. This involves building PostgreSQL + pgvector with debug symbols.

jkatz:
Per additional testing from @ankane, this may not be needed -- one can induce a parallel index scan without it (though currently it requires some core code hacking). I'd contend that "this is what I see the upstream index AMs doing" as one line for doing this, but it may not be convincing. I'll close this for now.

PR #169, opened 2023-06-29
Simplify TAP test structure to align with upstream

jkatz:
postgres/postgres@549ec20 moved to using "done_testing()" from the Perl testing framework, which removed the need to include test counts.


ankane:
Awesome, thanks!

Issue #163, opened 2023-06-15
Precision by Vector Dimension


olirice:
We've been digging into performance to improve documentation around required instance sizes, expected queries/sec, precision, etc

I had an unexpected result when measuring precision for different vector sizes and wanted to share to see if you have any insights

Holding the following parameters constant:

| parameter | value |
|-------|---|
| # records| 10_000 |
| # index lists | 30 |
| top k | 10 |
| metric | inner product |
| dataset | random unit vectors |

I varied the dataset's dimension and `ivfflat.probes` to measure the `precision@k=10`

That resulted in:
<img width="576" alt="Screenshot 2023-06-15 at 1 40 43 PM" src="https://github.com/pgvector/pgvector/assets/12958657/93d4c82f-fc58-46c9-81a0-05d83021620e">

*note: the legend was abbreviated, tested dimensions were `[3, 5, 10, 20, 30, 40, 50, ... 1400, 1600]`

Interpreting that chart:
As the number of probes approaches number of index lists, precision approaches 100% (as expected)

Another trend is that as the vector dimension grows, the relative benefit of the index decreases. That is visible as the curves "flatten out"

Another way to think about that is the amount of the dataset that has to be scanned to achieve a given precision.

For example:

<img width="576" alt="Screenshot 2023-06-15 at 1 17 09 PM" src="https://github.com/pgvector/pgvector/assets/12958657/dcf2ec0a-1e63-4b58-b756-993e2f5efd85">

so, to achieve a minimum of 99.0% precision on vector dimensions > 150, we have to compute distances for more than 80% of the records in the dataset.

---

Questions:

- Is that roughly what you expected?

- Is the characterization of the relationships among lists, probes, and vector dimensions (shown above) roughly correct, or am I missing some info?

- Do you think recommendations in the form of
"for approximately `M%` precision on `N` dimensional vectors, set `ivfflat.probes` = `Y%` of `lists` "
would be the right way to summarize for users?

jkatz:
@olirice This seems to be a pretty small data set, i.e. 10,000 rows, along with it being divided up over 30 lists. This is probably skewing the clusters a bit, so it's less likely you're going to find commonality without visiting multiple lists. At the size of this data set, even with higher dimensionality, I would suspect doing a sequential scan (not using the index) is likely the way to go if you're most interested in precision.

[...]

PR #178, opened 2023-07-05
Index build speed optimizations

pashkinelfe:
When looking at the low-level ivfflat build, I saw that two things take a significant share of time:

1. Calculation of dot product
2. Caclulation of arccos value.

I did some experiments and found that I could speed up these things with a marginal precision decrease (considering that overall ivfflat is approximate and the share of exact answers in select query is significantly < 100% at any reasonable ratio of probes/lists).

The code is as follows:
https://github.com/pashkinelfe/pgvector/tree/index-build-speed-optimizations

There are two patches

1. Use float instead of double at dot product calculation. It has the most pronounced effect as dot product calculation contains floating point multiplication of all vector dimensions. Intentionally I left double the overall distance function and the calculations that are done once per vector pair (not at every dimension) as they will increase speed only marginally (and also for compatibility). The low-level reason for this speed-up is that it makes CPU (armv8) using vector multiply-add instruction (fmadd) instead of vector multiplication + conversion to double + addition (fmul + fcvt + fadd) at each vector dimension. This change can also speed up the select's speed due to faster lists traversal that includes dot product calculation to the pairs of (sample vector - index vector).

2. Caclulate arccos value for spherical big circle distance as a quadratic Lagrange approximation plus sign extension. The effect of this is less pronounced so it could decided to be merged separately.

Overall performance measurements are done at original state (arccos + double dot product), first patch (arccos + float), second (Lagrange approximation + double) and both (Lagrange approximation + float). The dataset is a real 900K sized set of OpenAI vectors (https://huggingface.co/datasets/KShivendu/dbpedia-entities-openai-1M), but the same results are also for random 900K set. Number of lists are chosen at recommended value (900) and 3 time recommended and 1/3 times recommended. The most pronounced effect (55% index build time decrease) is when the number of lists is more than recommended, which has a good reason to be build to increase select speed at cost of index build time. But for recommended value the effect is also very pronounced i.e. 30% index build time decrease with both patches.

Absolute index build times:
image

Relative index build times as a ratio to original unpatched code:
image

Regarding index quality, I consider that these small changes in distance calculations will not be so pronounced to be seen on precision vs probes/lists plot as in https://github.com/pgvector/pgvector/issues/163#issuecomment-1603041726 or https://github.com/erikbern/ann-benchmarks. I don't expect seeing any changes in precision vs probes/lists or qps plot for selects performance. Still I'll try to publish these benchmarks for all patch variants vs current state.

### DIFF
--- a/src/vector.c
+++ b/src/vector.c
@@ -647,11 +647,9 @@ vector_spherical_distance(PG_FUNCTION_ARGS)
 	if (dp > 1)
 		dp = 1;
 
-	distance = negate ?
-		0.5 + (dp*(0.33333333*dp + 0.16666666)):
-		0.5 - (dp*(0.33333333*dp + 0.16666666));
+	distance = dp*(dp*(1.31196852822*dp - 1.45311472663) + 0.64114619624);
 
-	PG_RETURN_FLOAT8(distance);
+	PG_RETURN_FLOAT8(negate ? 0.5 + distance : 0.5 - distance);
 }
 
 /*
--- a/src/vector.c
+++ b/src/vector.c
@@ -632,6 +632,7 @@ vector_spherical_distance(PG_FUNCTION_ARGS)
 	Vector	   *b = PG_GETARG_VECTOR_P(1);
 	float		dp = 0.0;
 	double		distance;
+	bool 		negate;
 
 	CheckDims(a, b);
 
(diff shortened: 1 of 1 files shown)


jkatz:
+1 for https://github.com/pgvector/pgvector/pull/178/commits/c93807eab00b519b5f1751667a970b96eca2dc8a (moving to float4 for distances) at least for building; it feels odd to me to cast from `float4` to `float8` at the end given the loss of information from not storing the float8 values during the calculation phase, but I don't know if this negatively impacts the final results.

For https://github.com/pgvector/pgvector/pull/178/commits/422f0a2a4a1be571bc875b31d9964ab62ea5f63c I'd want to see what the changes are precision are, if any, against some of the known benchmarks (e.g. ANN Benchmarks as mentioned).

pashkinelfe:
@jkatz thanks for your review! I agree with you for https://github.com/pgvector/pgvector/commit/c93807eab00b519b5f1751667a970b96eca2dc8a and created a separate PR for this patch alone https://github.com/pgvector/pgvector/pull/180. Changing the distance functions interface from double to float will need modification of vector.sql interface which may need a major extension upgrade. If we're ready for this, I'd easily modify the patch as requested. As of now, I've left it as is for compatibility and a small expected build speed gain.

ankane:
Thanks again @pashkinelfe. Closing this out due to the findings in https://github.com/pgvector/pgvector/pull/180#issuecomment-1639895308.

PR #180, opened 2023-07-06
Speed up ivfflat build: use float instead of double for dot product

pashkinelfe:
This is the first patch of patchset https://github.com/pgvector/pgvector/pull/178 as requested in the comments https://github.com/pgvector/pgvector/pull/178#issuecomment-1622886656

Regarding the conversion of the distance functions interface from double to float it will need modification of vector.sql interface which may need a major extension upgrade. If we're ready for this, I'd easily modify the patch as requested in https://github.com/pgvector/pgvector/pull/178#issuecomment-1622886656 As of now I've left it as is for compatibility, and as this modifications will not make it much faster to build ivfflat index.

### DIFF
--- a/src/vector.c
+++ b/src/vector.c
@@ -504,8 +504,8 @@ l2_distance(PG_FUNCTION_ARGS)
 	Vector	   *b = PG_GETARG_VECTOR_P(1);
 	float	   *ax = a->x;
 	float	   *bx = b->x;
-	double		distance = 0.0;
-	double		diff;
+	float		distance = 0.0;
+	float		diff;
 
 	CheckDims(a, b);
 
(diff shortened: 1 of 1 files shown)


jkatz (approved):

ankane:
Thanks @pashkinelfe, this is really great!! (both the performance win and the benchmarks) 🚀 🚀 🚀 

Looks like it unlocks fused multiply-add on both aarch64 and x86_64 ([additional benchmarks](https://gist.github.com/ankane/96e9750405f0b89974edd7db01168f5d)).

We should probably apply it to the `cosine_distance` function as well.

ankane:
Updated `cosine_distance` in b710dc68a0fe1da291705321460639358b0662f8.

ankane:
Just fyi: switched to `double` for `vector_norm` in dab8f25d1c1fae4cbe1f1d4482f21d914b45f1bf to avoid overflows. It did not affect performance in my testing (as it's only called once per sample and once per tuple during index builds).

PR #189, opened 2023-07-20
Added bound option [skip ci]

pashkinelfe:
bound3 is a rebased bound2 on the current master. Some benchmarks:
1. Ann benchmark (added bound patch measurements to existing ones)
Bound patch almost entirely repeats pgvector-float patch from measurements as pgvector-float patch is already in the master branch, the current bound3 is based on.
image

3. Manual select speed test on 900K rows from 1M_openai_embeddings dataset with an increased number of probes = 250 (lists = 2000) 
`select id from emb order by vector <#> (select vector from emb_query order by random() limit 1) limit 10;`

Increased number of probes/lists make select scans sort more rows each time, and that's the scenario where we can gain from bounded sort. Actual number of lists in the scan would be 900000/2000*250 = 112500

The results of 20 selects for bound 0 (=no bound) and bound 10 in interleaved order.
image

Without bound some measurements in a series are significantly slower. Overall average time is 0.28s (bound 10) vs 0.34s (bound 0 = no bound). It's not so impressive gain and got at bigger than recommended number of lists. Though it's not worse and could be merged IMO. 

I guess are there conditions where the effect from bound sort would be more pronounced?

### DIFF
--- a/src/ivfflat.c
+++ b/src/ivfflat.c
@@ -14,6 +14,7 @@
 #endif
 
 int			ivfflat_probes;
+int			ivfflat_bound;
 static relopt_kind ivfflat_relopt_kind;
 
 /*
@@ -33,6 +34,10 @@ IvfflatInit(void)
 	DefineCustomIntVariable("ivfflat.probes", "Sets the number of probes",
 							"Valid range is 1..lists.", &ivfflat_probes,
 							1, 1, IVFFLAT_MAX_LISTS, PGC_USERSET, 0, NULL, NULL, NULL);
+
+	DefineCustomIntVariable("ivfflat.bound", "Sets the max results from index (experimental)",
+							NULL, &ivfflat_bound,
+							0, 0, INT_MAX, PGC_USERSET, 0, NULL, NULL, NULL);
 }
 
 /*
--- a/src/ivfflat.h
+++ b/src/ivfflat.h
@@ -83,6 +83,7 @@
 
 /* Variables */
 extern int	ivfflat_probes;
+extern int	ivfflat_bound;
 
 typedef struct VectorArrayData
 {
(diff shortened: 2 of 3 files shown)


jkatz:
I don't think the data supports adding this at this time, at least based on these tests. This gives users another knob for what seems to be a very minimal gain.

From my tests, the [`dynamic-pruning2`](https://github.com/pgvector/pgvector/tree/dynamic-pruning2) patch had more of a performance gain in terms of "boudning" the overall results, but it needs more study given the "final pruning" is passed to the executor and the index may not provide enough tuples for the executor to consider.

pashkinelfe:
My look into bounding sort was with a goal to get bounding conditions from plan actually, not just setting GUC before every select. Now we have index scan that doesn't know how many rows are requested, and modifying this isn't an easy task.  There are some proposals about sorting index scans in community. So this is to evaluate how much performance we could gain from sort bounding. From the results I suppose we don't gain much, I agree.

Merging of existing patch with GUC at least is not prohibited, it doesn't make things worse, but I don't insist.

jkatz:
@pashkinelfe We're saying similar things -- as I mentioned, the implementation in [`dynamic-pruning2`](https://github.com/pgvector/pgvector/tree/dynamic-pruning2) showed better gains, at least in my testing.

I thought about this a bit more: this could pair with [`dynamic-pruning2`](https://github.com/pgvector/pgvector/tree/dynamic-pruning2) , which does put an upper bound on the number of tuples that are retained during the scan, but currently the value is hardcoded. This is still tricky to get correct, because we don't know how many tuples we need to pass back to the executor during the scan in a way that would satisfy the limit (particularly due to visibility), and as such we may underrepresent the data set if we don't return enough tuples.

The primary advantage is that this would keep an upper bound on work memory, which would definitely help in performance in many cases, as folks are likely seeing their sorts spill to disk.

ankane:
Hi @pashkinelfe, thanks for benchmarking this. The minimal difference in speed-recall is consistent with when I've measured in the past. The difference in variance in interesting, but I agree with @jkatz that I don't think there's enough to justify a new setting.

PR #190, opened 2023-07-20
Modify guidance on troubleshooting index scans [skip ci]

jkatz:
This adds guidance on adjusting `random_page_cost` first for pgvector < 0.4.3, as we still want PostgreSQL to consider a sequential scan as part of the query plan, in case the sequential scan is more performant. This also discusses the tradeoffs of setting `enable_seqscan` to `0`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -331,7 +331,20 @@ You’ll need to use [dimensionality reduction](https://en.wikipedia.org/wiki/Di
 
 #### Why isn’t a query using an index?
 
-The cost estimation in pgvector < 0.4.3 does not always work well with the planner. You can encourage the planner to use an index for a query with:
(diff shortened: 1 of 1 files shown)


ankane:
Hi @jkatz, thanks for the PR. I think the current guidance will work for most situations, so I'd prefer to keep it simple.

jkatz:
@ankane I definitely agree with simplicity in general. However, if you're taking multiple actions within a transaction (such as in a stored procedure), you may be inadvertently forcing other queries to bypass a sequential scan that may be more performant. However, that would also be true with the `random_page_cost` guidance without further advice.

Given the updates in the newer releases and what should be their prevalence in the near term, I won't push hard on this one.

Issue #150, opened 2023-06-08
why vector_cosine_ops use function vector_negative_inner_product but not cosine_distance in vector.sql


left2right:
https://github.com/pgvector/pgvector/blob/master/sql/vector.sql
```
CREATE OPERATOR CLASS **vector_cosine_ops**
	FOR TYPE vector USING ivfflat AS
	OPERATOR 1 <=> (vector, vector) FOR ORDER BY float_ops,
	FUNCTION 1 **vector_negative_inner_product**(vector, vector),
	FUNCTION 2 vector_norm(vector),
	FUNCTION 3 vector_spherical_distance(vector, vector),
	FUNCTION 4 vector_norm(vector);
```

jkatz:
This is an optimization based on how the vectors are stored on disk.

[Cosine similarity](https://en.wikipedia.org/wiki/Cosine_similarity) is defined as the `inner product(A,B) / ||A|| * ||B||`. The cosine distance is 1 - cosine similarity.

However -- the vectors are normalized before they are stored in the index, which means they have a magnitude of 1. Therefore, the cosine distance becomes `1 - inner product(A,B)`. This saves the time having to perform the magnitude calculations and division, so it's an optimization.

ankane:
Thanks @jkatz

PR #201, opened 2023-07-29
fix bug : vector_cosine_ops

ikenchina:


jkatz:
@ikenchina This isn't a bug -- this is a performance optimization that pgvector can use due to the vectors in the index all being normalized. There are more details here:

https://github.com/pgvector/pgvector/issues/150#issuecomment-1583077877

ikenchina:
I got it. 
Thanks

PR #205, opened 2023-07-31
Enable pgvetor to support order by null

fanfuxiaoran:
Now, when the ORDER BY value is null, the scan with ivfflat index will return nothing, which is different with seq scan.

For instance:
SELECT * FROM t ORDER BY val <=> (SELECT NULL::vector);

### DIFF
--- a/src/ivfscan.c
+++ b/src/ivfscan.c
@@ -25,6 +25,17 @@ CompareLists(const pairingheap_node *a, const pairingheap_node *b, void *arg)
 	return 0;
 }
 
+static Datum
+getdistance(IvfflatScanOpaque so, Datum v1, Datum v2)
+{
+	/* Use procinfo from the index instead of scan key for performance */
+	if (DatumGetPointer(v1) == NULL || DatumGetPointer(v2) == NULL)
+		return Float8GetDatum(0.0);
+	else
+		return FunctionCall2Coll(so->procinfo, so->collation, v1, v2);
+}
+
+
 /*
  * Get lists and sort by distance
  */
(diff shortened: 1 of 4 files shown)


jkatz:
What is the use case for this?

Absent of better understanding that, I'm generally not inclined towards supporting this for several reasons:

1. `NULL` is not `0` , but the absence of data. If you need to perform this operation, You can do something like:

```sql
SELECT *
FROM t
ORDER BY val <=> COALESCE(NULL::vector, '[0,0,0...]'::vector);
```

2. In PostgreSQL, the K-NN GiST index support does not supporting finding the distance between a `NULL` value.

3. Absent a use case, this adds overhead to the primary distance operation, where the ultimate goal is to continue to speed it up.

fanfuxiaoran:
>  In PostgreSQL, the K-NN GiST index support does not supporting finding the distance between a `NULL` value.

Is it better to report an error like ` ivfflat index doesn't support finding the distance between a NULL value` instead of returning an empty result?

ankane:
Hey @fanfuxiaoran, thanks for the PR! It looks like returning results for NULL queries would be more consistent with other index types.

```sql
SET enable_seqscan = off;
CREATE TABLE items (embedding point);
INSERT INTO items (embedding) VALUES ('(0,0)');
CREATE INDEX ON items USING gist (embedding);
SELECT * FROM items ORDER BY embedding <-> NULL;
 embedding 
-----------
 (0,0)
(1 row)
```

Just pushed a [null-query branch](https://github.com/pgvector/pgvector/compare/null-query) that should fix it without impacting performance for non-NULL queries. Let me know if there are any thoughts.

---

Related: NULL values are *not* included in the index, which is different from other index types. However, I think it's fine to differ there, as only a subset of results are returned already.

fanfuxiaoran:
>Just pushed a [null-query branch](https://github.com/pgvector/pgvector/compare/null-query) that should fix it without impacting performance for non-NULL queries.

Need to call `GetDimensions` each time? which means reading the meta page every time. How about caching the NULL vector?  Personally, I prefer to save the `dimension` in `IvfflatOptions`, but I'm not sure if it's a good idea.

jkatz:
@ankane I like your implementation, I think this handles this case but with minimal overhead. Should we wrap the `NULL` cause with an `unlikely` call?

ankane:
Merged in the commit above (switched some of the tests to counts since the order isn't deterministic). Thanks for reporting and driving this @fanfuxiaoran!

Re unlikely: I don't think it'll make much difference since it's not a [very hot code path](https://github.com/postgres/postgres/blob/REL_15_3/src/include/c.h#L262-L266) (only called once per scan).

PR #222, opened 2023-08-10
Optimize the ivfflat cost estimate

fanfuxiaoran:
When table is toasted, assuming the index is also toasted, the 'numIndexPages' would be approximately equal to the 'rel->pages * ratio'

### DIFF
--- a/src/ivfflat.c
+++ b/src/ivfflat.c
@@ -114,14 +114,16 @@ ivfflatcostestimate(PlannerInfo *root, IndexPath *path, double loop_count,
 
 	get_tablespace_page_costs(path->indexinfo->reltablespace, NULL, &spc_seq_page_cost);
 
-	/* Adjust cost if needed since TOAST not included in seq scan cost */
+	/*
+	 * Adjust cost if needed since TOAST not included in seq scan cost
+	 * Assuming the index also uses table TOAST, the numIndexPages
+	 * would be approximately equal to the rel->pages * ratio. Other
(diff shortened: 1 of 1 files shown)


ankane:
Hi @fanfuxiaoran, thanks for another PR. The index isn't toasted, so I think this makes the cost estimate more inaccurate (lower than needed). Is there a specific situation you're optimizing for?

fanfuxiaoran:
>  Is there a specific situation you're optimizing for?

No, there is no specific case for the optimization.  I thought there should be one, but I cannot come up with one, as the cost model is too complicated. 

> The index isn't toasted, so I think this makes the cost estimate more inaccurate (lower than needed)

Thanks for your comments. You are right that this makes the cost estimate more inaccurate. But from my point of view, since sequence scan doesn't include the pages in the TOAST table, the index should also exclude that( pages in the toasted table) to make the planner choose the appropriate scan path.  While the estimated cost is far away from the actual cost, personally I think that is not a big problem as the cost model can make the planner do the right choice.

ankane:
I think it's better to minimize the inaccuracy (lower the cost just enough so it's better than a sequence scan), since other index types may not be inaccurate. There's a previous discussion in #133 if you haven't seen it already.

I appreciate the suggestion, but want to stay with the current approach.

jkatz:
For the archive, there may be some upstream PostgreSQL work on how we can improve cost estimation when performing a scan on a TOASTed column.

PR #233, opened 2023-08-19
Make pgvector headers available to others.

Florents-Tselai:
This makes  ` pgvector` headers available to other Postgres extensions, like ` #include "extension/vector/vector.h"`


ankane:
Hi @Florents-Tselai, thanks for the PR! I think we could do this for `vector.h` in an experimental capacity, but prefer to keep `ivfflat.h` private. Is there a reason to use `HEADERS_vector` instead of `HEADERS`?

Florents-Tselai:
@ankane I agree about keeping `ivvflat.h` private; I Removed It from the list.

From PG's relevant [docs](https://www.postgresql.org/docs/current/extend-pgxs.html) , since we've already defined `MODULE_big` above, the best practice should be using `HEADERS_vector` accordingly. 
But if you `make -n install`, you'll see they result in the same installation sequence.

ankane:
Thanks @Florents-Tselai. I don't see `HEADERS_$MODULE` mentioned as a best practice, so switched to `HEADERS` in a follow-up commit.

PR #250, opened 2023-09-01
Simplify code

pashkinelfe:
I tried to remove redundancies I found in the code and shorten it overall:
- Removed redundant variables
- Re-used code for similar distance functions

Pure refactoring, no changes in behavior.

### DIFF
--- a/src/vector.c
+++ b/src/vector.c
@@ -552,8 +552,6 @@ l2_distance(PG_FUNCTION_ARGS)
 {
 	Vector	   *a = PG_GETARG_VECTOR_P(0);
 	Vector	   *b = PG_GETARG_VECTOR_P(1);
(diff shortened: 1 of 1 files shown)


ankane:
Hi @pashkinelfe, thanks for the PR. However, this breaks auto-vectorization with MSVC, and I'd prefer to keep the functions independent as they are now.

pashkinelfe:
No problem. Which patch of these three does break auto-vectorization with MSVC?

ankane:
I know at least `400034a` does (you can [uncomment this](https://github.com/pgvector/pgvector/blob/0d86191eaf29431b252a4c25a771ff068457584a/Makefile.win#L20) to get the report).

PR #251, opened 2023-09-02
Ivfflat column dimension must be greater than 0

JoshInnis:
InitBuildState for the ivfflat index checks that the dimension of the column to be indexed has a dimension greater than or equal 0. However a vector cannot have a dimension of 0. 

Adjust the check in InitBuildState to follow the vector input logic which requires at least one dimension.


ankane:
Hi @JoshInnis, thanks for the PR. The `typmod` will never be 0, so this change won't have any effect.

PR #265, opened 2023-09-15
No need to MarkBufferDirty(); GenericXLogFinish() does that.

jeff-davis:


jkatz:
@jeff-davis Thanks for submitting! I believe this should reduce some overhead?

ankane:
Thanks @jeff-davis! Looks like the docs mention this as well.

> `GenericXLogFinish()` takes care of marking buffers dirty and setting their LSNs. You do not need to do this explicitly.

https://www.postgresql.org/docs/current/generic-wal.html

PR #321, opened 2023-10-27
Provide estimate of total tuples to insert on bulk HNSW build

### LINKED ISSUE #320: Indexing progress: tuples_total not populated while loading tuples
I want to create a HNSW index and followed the instructions in the readme. When I check the indexing progress I get something like this:

```sql
SELECT phase, tuples_done, tuples_total FROM pg_stat_progress_create_index;
             phase              | tuples_done | tuples_total 
--------------------------------+-------------+--------------
 building index: loading tuples |      394848 |            0
(1 row)
```

So `tuples_total` is not populated even if the phase is `loading tuples`. I understand from the readme that it should be populated, is this correct?

My pgvector version is 0.5.1.

jkatz:
We won't know how many tuples will actually be inserted until we finish the full table scan, but we can estimate the total based on the value avaulable in `pg_class.reltuples`. This allows for the index creator to get an approximate estimate of how long a build may take.

Note that this leaves the default value of `-1` for reltuples in place if there is no data available.

fixes #320

### DIFF
--- a/src/hnswbuild.c
+++ b/src/hnswbuild.c
@@ -20,6 +20,7 @@
 #include "access/tableam.h"
 #include "commands/progress.h"
 #else
+#define PROGRESS_CREATEIDX_TUPLES_TOTAL 0
 #define PROGRESS_CREATEIDX_TUPLES_DONE 0
 #endif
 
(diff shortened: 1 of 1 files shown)


ankane:
Thanks @jkatz. I think it's better to update the docs rather than use an incorrect value. Made small change in c599f92b52d20d1555654803f8f9a4955a97bc11, but it may be better to recommend:

```sql
SELECT phase, blocks_done, blocks_total FROM pg_stat_progress_create_index;
```

for HNSW.

jkatz:
@ankane 🤔 I actually want to check what we do for B-trees, because I would suspect that it's a similar issue.

jkatz:
@ankane Alright, nbtree uses exact:

https://github.com/postgres/postgres/blob/c6cf6d353c2865d82356ac86358622a101fde8ca/src/backend/access/nbtree/nbtsort.c#L496C2-L496C2

Tracing the code, the difference between the pgvector approach and nbtree is that nbtree pulls in the tuple data and puts it into spools for subsequent processing, which is why it's able to pull out the exact number of tuples to be processed. Based on the current implementation of HNSW, unless we do a full scan to get the exact count at the time, we at best could estimate.

If you're not inclined to provide an estimated value (which could be accurate at the given time), then I'm OK to close this and instead provide a doc update.

ankane:
Updated the docs with a better query in the commit above (which will also be accurate for partial indexes).

jkatz:
@ankane Looks great -- thanks!

Issue #33, opened 2022-08-24
Q: Index Creation


atanastrpceski:
I have a question about index creation. 

What happens if you don't specify the index operation type

So instead of doing it like this:

```
CREATE INDEX ON items USING ivfflat (embedding vector_l2_ops);
CREATE INDEX ON items USING ivfflat (embedding vector_ip_ops);
CREATE INDEX ON items USING ivfflat (embedding vector_cosine_ops);
```

I just do:

`CREATE INDEX ON items USING ivfflat (embedding); `  (it does allow me to do this)

Am I getting something if I do it like this?

What will be the approach if I like to use all 3 operations in my app (l2, ip, cosine), should I create 3 indexes?

ankane:
Hey @atanastrpceski, the default opclass is `vector_l2_ops` if one isn't specified (in hindsight, I wish I hadn't set a default to avoid confusion and may try to remove it in the future, but it's a breaking change).

And exactly - you'll want to create 3 indexes if you want to index all 3 operations.

atanastrpceski:
Thanks!

PR #342, opened 2023-11-06
Make the vector_l2_ops to be hnsw's default operator

fanfuxiaoran:


ankane:
Hi @fanfuxiaoran, thanks for another PR, but this was an intentional decision (see #33 for more context).

ankane:
I think it isn't clear which operator you're indexing (like in the thread), and could be interpreted as indexing all of them. Neither is an issue when requiring the opclass.

PR #346, opened 2023-11-08
Update docker image to postgres 16

stefansundin:
Hello, I'd like to use the pgvector docker image with postgres 16.

I built this and did some cursory testing and things seem to work ok. I did not do any extensive testing yet though.

Please let me know if you'd like me to change anything.


ankane:
Hi @stefansundin, thanks for the PR. You can build the image manually with Postgres 16 from the instructions in the readme, but I don't want to update the default until 0.6.0.

I'm also planning to start publishing an image for each major version of Postgres to make things easier for users (starting with 0.6.0).

PR #352, opened 2023-11-12
Add docker publish workflow

manukall:
This adds a Github Workflow to build and publish a Docker image after every push to master. The images would be available as `ghcr.io/pgvector/pgvector:XX`, where XX is the Postgres major version. 

If you think this is a good idea, we should probably also adapt the Readme and point to these images instead.

I did not include Postgres versions 11 and 17, because building didn't work on 11, and there is no `postgres` Docker image for 17 yet.

### DIFF
--- a/.github/workflows/release_docker_image.yml
+++ b/.github/workflows/release_docker_image.yml
@@ -17,13 +17,11 @@ jobs:
       fail-fast: false
       matrix:
         include:
-        - postgres-major: 11
         - postgres-major: 12
         - postgres-major: 13
         - postgres-major: 14
         - postgres-major: 15
         - postgres-major: 16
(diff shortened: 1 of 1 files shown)


ankane:
Hi @manukall, thanks for the PR. I'm planning to start publishing an image for each major version of Postgres starting with 0.6.0, but only for releases, and want to continue using Docker Hub (still figuring out the details, so not sure if it'll be a workflow in this repo or something else).

manukall:
Hi @ankane. That sounds like a good plan. I basically created this for myself, because I wanted to easily use Postgres 16 with pgvector in a Github service container for CI.

Thanks for your work on this awesome extension!

PR #360, opened 2023-11-16
Allow specify dimensions in CREATE INDEX

fanfuxiaoran:
As the typmod of vector can be -1, which means that we can create a table and insert data into it without caring the dimension. For ivfflat and hnsw index, they both need the dimension, so I think we can set the dimension when creating index.

In some cases, dimensions of vector cannot be determined prior to seeing the vector. For example, the vectors might be generated by a complex UDF. Such UDFs are possibly developed by data scientists and the details might be hard to comprehend for the user calling them.

This means users who want to load the generated vectors and do the search when developing the application will not be able to specify the dimensions of the vectors before the vectors are written to the table and the type of the column will only be vector instead of vector(<dims>). As a result, when creating index, there will be an error like

ERROR:  column does not have dimensions
To mitigate this issue, this patch adds an option dims to the access methods, allowing the user to specify the dimensions after seeing the vectors in the table. In this way, the user does not need to care about how the vectors are generated in order to search them.

### DIFF
--- a/src/hnsw.c
+++ b/src/hnsw.c
@@ -37,6 +37,10 @@ HnswInit(void)
 #endif
 		);
 
+	add_int_reloption(hnsw_relopt_kind, "dims", "Expected number of vector dimensions",
+					  -1, -1, HNSW_MAX_DIM
+					  ,AccessExclusiveLock
+		);
 	DefineCustomIntVariable("hnsw.ef_search", "Sets the size of the dynamic candidate list for search",
 							"Valid range is 1..1000.", &hnsw_ef_search,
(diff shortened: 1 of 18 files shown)


ankane:
Hi @fanfuxiaoran, thanks for another PR. You can currently index a `vector` column without dimensions with [this approach](https://github.com/pgvector/pgvector#can-i-store-vectors-with-different-dimensions-in-the-same-column). Support for indexing Postgres arrays will likely require a `dimensions` option in the future (like in the [hnsw-datum branch](https://github.com/pgvector/pgvector/compare/hnsw-datum)), but I'd like to hold off until then.

PR #362, opened 2023-11-20
Support building without PGXS

japinli:
Currently, pgvector only support build with PGXS. This commit tries to support non-PGXS building.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -28,19 +28,19 @@ jobs:
         with:
           postgres-version: ${{ matrix.postgres }}
           dev-files: true
-      - run: make
+      - run: make USE_PGXS=1
         env:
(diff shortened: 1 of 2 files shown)


japinli:
See the check logs, I find the workflow is:

``` 
     - run: |
          export PG_CONFIG=`which pg_config`
          sudo --preserve-env=PG_CONFIG make install
      - run: make installcheck
```

Maybe we should update the workflow to use PGXS.

```
     - run: |
          export PG_CONFIG=`which pg_config`
          sudo --preserve-env=PG_CONFIG make USE_PGXS=1 install
      - run: make USE_PGXS=1 installcheck
```

ankane:
Hi @japinli, thanks for the PR, but still have the same thoughts as #18 and #98.

PR #371, opened 2023-12-02
Use LockBuffer instead of LockBufferForCleanup when doing bulkdelete

fanfuxiaoran:
As both ivfflat and hnsw don't support non-mvcc snapshot, so there is no need to call LockBufferForCleanup when doing bulkdelete.

### DIFF
--- a/src/hnswvacuum.c
+++ b/src/hnswvacuum.c
@@ -463,13 +463,7 @@ MarkDeleted(HnswVacuumState * vacuumstate)
 
 		buf = ReadBufferExtended(index, MAIN_FORKNUM, blkno, RBM_NORMAL, bas);
 
-		/*
-		 * ambulkdelete cannot delete entries from pages that are pinned by
-		 * other backends
-		 *
-		 * https://www.postgresql.org/docs/current/index-locking.html
-		 */
-		LockBufferForCleanup(buf);
+		LockBuffer(buf, BUFFER_LOCK_EXCLUSIVE);
 
 		state = GenericXLogStart(index);
 		page = GenericXLogRegisterBuffer(state, buf, 0);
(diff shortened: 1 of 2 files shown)


hlinnaka (approved):
Yep, seems correct to me.

Furthermore, pgvector never holds a pin on a page after unlocking it. The only exception is during HNSW index build in HnswBuildAppendPage(), which briefly unlocks a page. That's OK because it doesn't hold any pointers to the page contents, and also because vacuum cannot run while the index is being built.

fanfuxiaoran:
Thanks for your review. 

> Furthermore, pgvector never holds a pin on a page after unlocking it. The only exception is during HNSW index build in HnswBuildAppendPage(), which briefly unlocks a page. That's OK because it doesn't hold any pointers to the page contents, and also because vacuum cannot run while the index is being built.

Yes,  in HnswBuildAppendPage(), it reads the new buffer and locks it, then sets the  `nextblkno` of the old page to it, and then unlocks the new buffer. Is it necessary to lock the new buffer as it only gets its page number?

ankane:
Hi @fanfuxiaoran, thanks for another PR. However, I don't see a strong reason to change this right now unless there's a noticeable difference in performance.

fanfuxiaoran:
>However, I don't see a strong reason to change this right now unless there's a noticeable difference in performance.

fanfuxiaoran:
> Hi @fanfuxiaoran, thanks for another PR. However, I don't see a strong reason to change this right now unless there's a noticeable difference in performance.

Yes, I can understand your point. This pr cannot improve the performance significantly. The overhead of waiting for the pin-count to be 1 is not big, I have tested that, and most of the time, the pin-count is 1.  Personally, I just think "LockBuffer" is more simple.

ankane:
Going to keep the current approach for now, but may revisit in the future.

Issue #120, opened 2023-05-09
Indexing vectors have size 2048


koseemre:
Hi. I am trying to index vectors with size 2048.

Firstly I build alpine-based postgresql version 15 with configs below
		--with-blocksize=32 \
		--with-wal-blocksize=32 \

Then build the pgvector image using build postgresql base with the updated IVFFLAT_MAX_DIM value in ivfflat.h.

Finally the block_size seems 32768 in DB, but when I tried to index my vectors, DB gives an error below:

**ERROR: index row requires 8208 bytes, maximum size is 8191.** 

Is there any suggestion about the case ?

postgresql build ref: https://github.com/docker-library/postgres/tree/master/15/alpine

pgvector alpine base build ref: https://github.com/pgvector/pgvector/pull/4

ankane:
Hi @koseemre, thanks for reporting! It looks like the max index tuple size is [hard-coded](https://github.com/postgres/postgres/blob/c5b7f67fcc8c4a01c82660eb0996a3c697fac283/src/include/access/itup.h#L65) as 8191 bytes in Postgres, so the suggestion in the readme won't work. I've removed it in the commit above. The one other option at the moment is dimensionality reduction.

nreith:
@ankane Does this change in the commit you mentioned have any affect on max dimensions of 16k, or max indexed dimensions of 2k?

montanalow:
There is a report that this can be changed pretty easily upstream in PG without significant repercussions, although I've not tested it. Has anyone from pgvector reached out to the core team to see if they're open to a patch to make this configurable similar to blocksize? 

https://dba.stackexchange.com/questions/244624/how-to-increase-index-row-maximum-size-in-postgresql

jkatz:
@montanalow The comment there says "it might have repercussions in other parts of the code" and would warrant further investigation. It may be worth reaching out to -hackers to understand what the ramifications are of changing it.

montanalow:
I've submitted a patch to -hackers to start a discussion on increasing the limit so we can understand the ramifications.

PR #402, opened 2024-01-09
upgrade to 0.5.2; Increased max dimensions for index from 2000 to 4096

MarchLiu:
# Increased max dimensions for index from 2000 to 4096

[Ollama generate embedding vector](https://github.com/jmorganca/ollama/blob/main/docs/api.md#generate-embeddings) with 4096 dimensions. So I increase the constant.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.5.2
+
+- Increased max dimensions for index from 2000 to 4096
+
 ## 0.5.1 (2023-10-10)
 
 - Improved performance of HNSW index builds
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 EXTENSION = vector
-EXTVERSION = 0.5.1
+EXTVERSION = 0.5.2
 
 MODULE_big = vector
 DATA = $(wildcard sql/*--*.sql)
@@ -63,7 +63,7 @@ prove_installcheck:
 
 dist:
 	mkdir -p dist
-	git archive --format zip --prefix=$(EXTENSION)-$(EXTVERSION)/ --output dist/$(EXTENSION)-$(EXTVERSION).zip master
+	git archive --format zip --prefix=$(EXTENSION)-$(EXTVERSION)/ --output dist/$(EXTENSION)-$(EXTVERSION).zip v0.5.2
 
 .PHONY: docker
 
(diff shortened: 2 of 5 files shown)


jkatz:
This PR will fail without additional changes. Currently, you can only index vectors up to 2K dimensions due to the PostgreSQL page size being 8KB by default. You can increase this limit if you recompile PostgreSQL with a larger page size.

However, should the work from https://github.com/pgvector/pgvector/tree/tinyint and/or https://github.com/pgvector/pgvector/tree/half be merged in, this will allow increasing the indexable vector size for those dimension sizes.

ankane:
Increasing the block size won't help either (#120), so the one option right now is dimensionality reduction.

MichaelKarpe:
> [Ollama generate embedding vector](https://github.com/jmorganca/ollama/blob/main/docs/api.md#generate-embeddings) with 4096 dimensions. So I increase the constant.

Nice try! :smile:

MichaelKarpe:
> Increasing the block size won't help either (#120), so the one option right now is dimensionality reduction.

@ankane @jkatz for my understanding, does it mean that it's not technically possible to increase to more than 2000 dimensions in pgvector or that it requires significant developments?

Most of the time requests on dimension size increase for indexation come for avoiding the dimensionality reduction, but there may be use cases where that many dimensions are necessary for ensuring a minimum search relevance. Thus, it'd be more than welcome to have greater limit as for many competitors.

For instance, would you know how pgvecto.rs is able to get the vector dimension limit for indexation to 65,535? See [their comparison with pgvector.](https://github.com/tensorchord/pgvecto.rs?tab=readme-ov-file#comparison-with-pgvector).

> Hi again, other than PCA is there a method to create indexes on vectors more than 2000? I have 2048 ones, right on the limit :(. If PCA is the only solution, what are some outcomes, like does it efficiently work?

@TutubanaS with pgvector it's currently not possible to go above 2000 dimensions AND having the indexation. However, it's possible to go up to 16,000 dimensions in pgvector, but without indexing chances are very high that the query per second (QPS) performance will be very bad and not reasonable for a business production use case.

Dimensionality reduction may (will most probably if you're using some pretrained model) create some loss of information and deteriorate the relevance of your search results. It's hard to evaluate how much the performance may deteriorate without more information. I'd recommend you perform the dimensionality reduction and evaluate the search relevance with and without it. Most probably here the quality loss will be very small since the dimensionality reduction is small from 2048 to 2000 dimensions.

jkatz:
> get the vector dimension limit for indexation to 65,535

When I last checked, this is due to not using the PostgreSQL storage system. The data is stored in an independent format, but the tradeoff is that does not use PostgreSQL's durability mechanisms.

PR #404, opened 2024-01-09
Add bullet points to Indexing section

danthedaniel:
This makes it more obvious what differences exist between IVFFlat and HNSW

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,7 +162,12 @@ You can add an index to use approximate nearest neighbor search, which trades so
 Supported index types are:
 
 - [HNSW](#hnsw) - added in 0.5.0
+  - Better query performance than IVFFlat
+  - Slower build times than IVFFlat
+  - Uses more memory than IVFFlat
+  - Does not need to train on seed data
 - [IVFFlat](#ivfflat)
+  - Requires pre-existing seed data to train the index
 
 ## HNSW
 


README.md
```diff
@@ -162,7 +162,12 @@ You can add an index to use approximate nearest neighbor search, which trades so
 Supported index types are:
 
 - [HNSW](#hnsw) - added in 0.5.0
+  - Better query performance than IVFFlat
```
jkatz, line 165: This isn't necessarily true. Based on how you construct and query your IVFFLAT index, pure query performance could be faster than HNSW. Where HNSW excels is that it typically has a better performance/recall ratio than IVFFLAT.

jkatz (commented):
Thanks for the submission! I think it's a good goal to provide guidance on how to choose an index. However, the topic is a bit more nuanced, and we should cover things like data ingestion patterns (is the data being bulk loaded, or iteratively updated), etfc.

README.md
xzilla: Yeah, I feel like these bullet points would be better served as a dedicated "HMSW vs IVFFlat" section (or maybe in the FAQ).

danthedaniel: @jkatz - I pulled it from the README's existing text:

> It has better query performance than IVFFlat (in terms of speed-recall tradeoff), but has slower build times and uses more memory.

ankane:
Hi @danthedaniel, thanks for the PR. I'm not really a fan of this change since the information is repeated directly below, but will continue to think and iterate on this.

danthedaniel:
Okay no worries

PR #422, opened 2024-01-20
Add FAQ HNSW vs IVFFlat?

xzilla:
Summarize the trade-offs at play when choosing between HNSW or IVFFlat indexes.

Inspired by recent PR #404 by @danthedaniel.


gwenshap:
This looks good and summarizes well what I've found in other sources. Merging it will be helpful.

michaelbautin:
I also think it would be useful to merge this.

PR #437, opened 2024-01-26
Mention setting toast_tuple_target in README.md

pashkinelfe:


ankane:
Hi @pashkinelfe, please see https://github.com/pgvector/pgvector/issues/409#issuecomment-1909064092 (changing the storage for the column is a more targeted way to do this, and it'd be good to have benchmarks before adding to the readme). Also, this can cause rows to become too wide, causing inserts to fail.

pashkinelfe:
Changing toast_tuple_target for a table from to 8Kb allows not to toast the rows between 2Kb and 8Kb. If they are larger they need to be toasted anyway due to page size. At least this will not cause insert errors unllike storage PLAIN. I'd try to do benchmarking some time.

ankane:
Let's revisit when there are benchmarks.

Issue #27, opened 2022-05-25
Ideas


ankane:
Please create a new issue to discuss any ideas or share your own.

Plan

- [ ] Use pairing heap for index scan for performance - `stages` branch
- [ ] Use mini-batch k-means for index creation for reduced memory - `minibatch` branch
- [ ] Add support for product quantization (in-progress)

Ideas

- [ ] Use `tuplesort_set_bound` for performance - `bound` branch (not needed w/ pairing heap)
- [ ] Add functions to view lists and/or pages like [pageinspect](https://www.postgresql.org/docs/current/pageinspect.html) (require superuser)

On-hold

- [ ] Add support for parallel index scans (planner gets cost estimate but doesn't use) - `parallel-index-scan` branch
- [ ] Change return type of distance functions from float8 to float4 for performance (maybe, needs benchmarking)

PR #463, opened 2024-02-17
Offer pre-built `pgvector` releases

tachyonicbytes:
First of all, thanks for the great work on `pgvector`.

I have been working on a `pgvector` powered project in the recent past, and one of the problems we encountered was building `pgvector`. We needed only the built files to place into a custom `postgres` installation (that lacked `pg_config`, for example), and even though there are many ways to run `pgvector`, it seemed hard to get just the build files.

I wanted to add a github action that builds `pgvector` for a suitable matrix of `postgres` versions and operating systems. But it turns out that `build.yml` already does that, it just doesn't keep the built files around.

This PR adds the build-artifact github action to `build.yml` jobs.

This is just a simple example, if this direction is approved by the maintainers, we should also add the built files to the proper release. For now, they are available on the actual github action run, as artifacts.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -91,8 +91,7 @@ jobs:
           call "C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Auxiliary\Build\vcvars64.bat" && ^
           nmake /NOLOGO /F Makefile.win && ^
           nmake /NOLOGO /F Makefile.win install && ^
-          nmake /NOLOGO /F Makefile.win installcheck
-      - run: |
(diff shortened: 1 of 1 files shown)


ankane:
Hi @tachyonicbytes, thanks for the suggestion / PR.

There are already pre-built releases for a number of installation methods (Docker, Homebrew, APT, Yum, pkg, conda-forge, and Postgres.app), and I'm hoping to add an installer for Windows at some point (#27 - have this partially done in a private repo).

I don't think Postgres installations without `pg_config` are common, and manually copying files into a few different directories seems error-prone, so it's not something I'd like to support (also, it needs to include all `.sql` files for upgrades, fwiw).

PR #472, opened 2024-02-26
Add bulk-loading capability to README

oneturkmen:
Inserting vectors via `INSERT INTO ... VALUES` might be much more expensive. Instead, `COPY` is preferred due to being much faster:

```sql
COPY items (embedding) FROM STDIN
```

image

Reference: 
- https://dev.to/josethz00/speed-up-your-postgresql-bulk-inserts-with-copy-40pk
- https://stackoverflow.com/questions/46715354/how-does-copy-work-and-why-is-it-so-much-faster-than-insert


ankane:
Great idea, thanks @oneturkmen!

PR #506, opened 2024-03-29
[Index] Bumping HNSW + IVFFLAT max dimensions to 4096

philipk19238:


jkatz:
This won't work, as PostgreSQL index pages are limited to 8KB in size, and increasing the max dimension in this way without modifying how the vectors will cause PostgreSQL to fail to save the data (and I believe crash).

The [halfvec](https://github.com/pgvector/pgvector/tree/halfvec) branch shrinks the amount of data held in each dimension, which allows for an increase in the total number of dimensions that can be stored in an index (in the case of this branch, it would be 4K). @ankane has a few other branches open that are working with other different types of quantization to allow storing more dimensions of a vector in an index.

Issue #508, opened 2024-04-01
0.7.0 Plan


ankane:
Hi all, just wanted to share the initial plan for 0.7.0, with a goal for the end of April. If you want to help test or have any thoughts, check out the open issues, or create a new issue if there isn't one.

HNSW indexing support for more types

- float16 vectors - `halfvec` type - [early results](https://github.com/pgvector/pgvector/issues/326#issuecomment-2028467121)
- binary vectors - `bit` type (already in Postgres)
- sparse vectors - `sparsevec` type

New functions (all of these are currently possible with SQL but less convenient)

- `binary_quantize` - convert a vector into a binary vector
- `l2_normalize` - normalize a vector to length 1
- `subvector` - slice a vector

New operators

- Hamming distance - `<~>`
- Jaccard distance - `<%>`
- L1 distance - `<+>`
- concatenate vectors - `||`

Other changes

- HNSW <strike>and IVFFlat</strike> indexing support for L1 distance
- dynamic dispatch for Linux
- change comparison operators to not throw an error for vectors with different dimensions
- change handling of range error to be consistent with `real`

ankane:
Quick update: 0.7.0 is on-track to be released early next week (pending final testing).

ankane:
0.7.0 is now released 🎉

PR #524, opened 2024-04-18
Add search option to better process queries with WHERE clause (relaxed monotonicity)

VladlenPopolitov:
The current state: pgvector uses HNSW algorithm, that returns k-TOP nearest neighbours. If SQL query has WHERE clause, k-TOP neighbours could not satisfy the filter condition, and the query does not return any records. 

The proposal: I propose to add an option in HNSW search to use the relaxed monotonicity proposed in the article https://www.usenix.org/system/files/osdi23-zhang-qianxi_1.pdf . The implementation does not affect the index build, it changes the greedy search algorithm only. If SQL query does not have WHERE clause, user could use standard greedy search (less overhead, more quick). If SQL query has WHERE clause (and even does not have), the user could use relaxed monotonicity algorithm (higher recall guaranteed). This choice is implemented by GUC variable easily. 

In the proposed pull request the option  is activated by setting `SET hnsw.use_relaxed=ON` . When this option is activated, HNSW returns nearest vectors until needed records are found using Open-Next interface, instead of efSearch-TOP records in the standard implementation. If `hnsw.use_relaxed=OFF` (default), HNSW behaves as usual.

Changes: 
hnsw.c: PostgreSQL variable `hnsw.use_relaxed` and corresponding C variable `bool hnsw_use_relaxed` are added.
hnsw.h: Open-Next interface requires to store the search state between gettuple() calls, this state is stored in the members fields of the `struct HnswScanOpaqueData`
hnswscan.c: gettuple() was modified. If `hnsw.use_relaxed=ON` , it calls function OpenScanItemsRelaxed() instead of GetScanItems(), then it call NextScanItemsRelaxed() to get a next tuple.
hnswutils.c : HnswSearchLayerRelaxedNext(), HnswVisitedRelaxedNext() , HnswSearchLayerRelaxed() , NextScanItemsRelaxed() are added to implement Open-Next interface for greedy search.  
tests/sql/hnsw_relaxed.sql and tests/expected/hnsw_relaxed.out added to check the work of HNSW search with and without hnsw.use_relaxed option. These files contain also the tests from hnsw_vector_l2 and hnsw_vector_ip to ensure the standard test execution, when the option in ON.
README.md was NOT updated until final commit (GUC variable names could be changed or added, comments are expected) 

The example to demonstrate, that pgvector can work with WHERE condition using hnsw.use_relaxed option:
```SQL
SELECT t, 
Array[t,t+1,t+2]::vector(3) AS val 
INTO test_where 
FROM generate_series(1, 2000) as t;
SELECT t FROM test_where WHERE t>1500 ORDER BY val <-> '[1000.2,1000,1000.1]' LIMIT 2;
  t   
------
 1501
 1502
(2 rows)

CREATE INDEX ON test_where USING hnsw (val vector_l2_ops) WITH (m=10);
set hnsw.use_relaxed=off;
set hnsw.ef_search=100;
SELECT t FROM test_where WHERE t>1500 ORDER BY val <-> '[1000.2,1000,1000.1]' LIMIT 2;
 t 
---
(0 rows)

set hnsw.use_relaxed=on;
SELECT t FROM test_where WHERE t>1500 ORDER BY val <-> '[1000.2,1000,1000.1]' LIMIT 2;
  t   
------
 1501
 1502
(2 rows)
```
ANN-benchmark without WHERE (see picture) condition in query shows the decrease the query per seconds rate (additional overhead to support Next interface), that in many cases compensated by increase of the recall (more vectors in the index are checked). Taking into consideration WHERE condition, pgvector with hnsw.use_relaxed option has superior recall comparing to standard HNSW implementation of all programs.

I am open for comments and discussions regarding this pull request.

Vladlen Popolitov.

pgvector_standard_vs_relaxed_lastfm


jkatz:
> ANN-benchmark without WHERE (see picture) condition in query shows the decrease the query per seconds rate (additional overhead to support Next interface), that in many cases compensated by increase of the recall (more vectors in the index are checked). Taking into consideration WHERE condition, pgvector with hnsw.use_relaxed option has superior recall comparing to standard HNSW implementation of all programs.

What dataset(s) was used to test the above? What were the PostgreSQL configurations and instance that it was run on.

Did you test the recall/performance on a dataset that specifically was testing the filters?

VladlenPopolitov:
1. The reason to make ANN-benchmark without WHERE is to ensure, that there is no significant performance degradation, that can kill this idea. I tested this idea on hnswlib library and did benchmark, that had similar results. After that I implemented it in pgvector code and checked in benchmark. Attached picture above has results from benchmark with lastfm dataset (run on MacBookPro M1). PostgreSQL 16.1, maintanance_work_mem 16Gb, shared_buffers 4Gb. lastfm data set on the server with AMD EPYC 9654(AVX256) has the similar results. Also I tested on datasets glove-100 and fashion-mnist: decrease of rate 0-10% with the increase of recall by 0-3% (as value). The increase of a recall is mentioned in the article.
2. Benchmark with WHERE was not used. The article discuss the dataset Recipes1M+, that needs the permission for usage, that is not suitable for open-source benchmarks. I think now, how to prepare standard datasets for benchmarks with WHERE based on the standard datasets from ANN-benchmarks (with equal, less, more conditions in WHERE clause). The benchmark with WHERE is relevant only for algorithms , that implement filtering. Algorithms without filtering will show very low recall. It is like 2 different generations, it is not fair to compare them.

ankane:
Hi @VladlenPopolitov, thanks for the PR! I'm planning to focus on inline filtering after 0.7.0 (#508), so should have more thoughts then (it may take some time, so patience is appreciated). Iterative searching (like this or #282) seems promising, but I also want to explore other options.

jonahharris:
@ankane @jkatz is there any reason this wasn't picked up? As @VladlenPopolitov mentioned, this performs substantially better when predicates are present, which is a pretty common case I've seen. I have a slightly cleaned-up version, but wanted to see what the status of this is before doing anything with it.

jkatz:
@jonahharris We've collectively been doing a lot of research on filtering, and it's a very nuanced topic. I'm hoping to share my own findings soon.

jonahharris:
Cool. Moving closer to VBASE's approach would be good, IMO - but maybe there is newer research (aside from the vector/scalar hybrid variants directly incorporating scalars). Of the PG vector index implementations I've seen so far, none seem to do proper costing/selectivity in the filtering case.

VladlenPopolitov:
Hi @ankane ! I did benchmarks for pgvector in standard mode and relaxed mode and found, that search results depends more on filters value distributrion, than search algorithms: there is no universal best search algorithm.
1) When there is no filter, standard pgvector slightly quicker, than relaxed (due to relaxed's overhead to store all visited vectors- I showed it on picture in pull request comment above) 
2) When there is filter condition with ~2-10 values, relaxed algorithm becomes faster than standard (relaxed decreases speed with the remaining recall, standard increases efsearch to keep the same recall resulting higher speed decrease)
benchmark3_1

3) When filter has ~100 values relaxed drops speed keeping recall rate, standard drops recall to 1-10% and becomes unusable, it returns empty datasets instead of some not nearest vectors.
benchmark3_1

4) When filter has ~1000 values, standard returns almost always empty datasets, relaxed returns the same recall, but slower than table scan without index. In this situation usual btree index works very fast (quick filter and then distance calculation for filtered tuples from table).
benchmark3_1

5) When filter has ~1000 values and table is very big, then btree index also does not work. In this situation composite index can help, that naturally process filter conditions.

relaxed mode gives to users more options to search, when they need filters. Also I would point, that pgvectorscale also implemented similar algorithm, when results returned until filter conditions are satisfied and all records returned.
I think, maybe "relaxed" name is not properly explain the method, probably better is "stream".

I could update pull request to the current code base. It is also interesting, what are your findings regarding filtered search.

heathiviesyncro:
Hello: Is it possible to get this branch updated to resolve the conflicts?

VladlenPopolitov:
@heathiviesyncro  pull request updated to the current master version.
I did benchmark to check the correct work (in addition to tests), it works on the same performance, but has post-filter capabilities as additional feature.
benchmark3_1

ankane:
Hi @VladlenPopolitov, thanks for sharing the benchmarks, and sorry for not getting to this sooner.

I've been spending time on this area recently and think I've found an approach that works well with vacuuming (#678). Let me know if you have any feedback.

heathiviesyncro:
> @heathiviesyncro pull request updated to the current master version. I did benchmark to check the correct work (in addition to tests), it works on the same performance, but has post-filter capabilities as additional feature. benchmark3_1

Thank you very much for getting this updated.

ankane:
0.8.0 is now released, so closing this out.

Big thanks @VladlenPopolitov for helping move this feature forward (and finding the `TRUNCATE` crash).

Issue #326, opened 2023-10-29
Can I help get tinyint or half branches released?


nathanwilk7:
Is there more to do on the `tinyint` or `half` branches to get them released or are they ready to be put into `0.5.1`? If there is more to do for them, let me know and I'll see if it's something I could take care of (e.g.: code, docs, tests, etc).

ankane:
Hi @nathanwilk7, the most helpful thing would be sharing specific use cases for either data type (how the vectors are generated and what specifically they're used for). These may be included in 0.6.0, but I'd like to understand how common they will be.

nathanwilk7:
Sure, I'll add my use cases here. For context, we're doing chem/bio ML type work. Thanks for your thoughts on the below.

Our smaller datasets have about 300 million (300M) molecules in them. For those molecules there are a few different types of vectors we'd like to generate. Some of these vectors are generated via cheminformatics methods (basically a molecular hash function with certain similarity properties) and others are generated via embeddings from various ML models.

1. 300M sparse "count" vectors w/ 1024-2048 dimensions ([Morgan Count Fingerprints](https://greglandrum.github.io/rdkit-blog/posts/2023-01-18-fingerprint-generator-tutorial.html)]. Most of the dimensions will be 0, but the non-zero ones will be integers between 0 and some small N (e.g.: 32 would likely be a reasonable upper bound, 256 would be extremely high). For these vectors, we'd like to be able to do ANN searches across them using the standard distance measures (inner product, cosine, L2). The goal here is to able to run searches to find the most similar molecules to a given molecule or molecule fragment. The tinyint type seems like it could be a good fit here, though supporting spare vectors would potentially be an even bigger win but using dense tinyint might be good enough for now.
2. 300M embedding vectors w/ 128-1024 dimensions where each dimension is a non-zero decimal number (these are essentially the same as the standard embeddings everyone uses for various ML tasks). We would likely be ok giving up the precision of using half size floats or product quantization or any other similar technique.
3. I'll also add that we'd like 300M sparse bit vectors w/ 1024-2048 dimensions ([Morgan Bit Fingerprints](https://www.rdkit.org/docs/GettingStartedInPython.html#fingerprinting-and-molecular-similarity)]. For these vectors, we'd like to be able to do ANN searches across them using tanimoto/jaccard distance. I recognize these are probably not going to be supported as true ANN-searchable sparse bitvectors in pgvector anytime soon, so we can probably ignore this unless you think it's on the roadmap.

[...]

PR #536, opened 2024-04-30
SVE vector optimization for halfvectors dot product calculation

pashkinelfe:
Hi, @ankane and @jkatz !

I optimized the dot product calculation for half vector using SVE extension that is present on the many machines suitable for vector search on ARM architecture.
 
1. Testing HNSW build time (same test as https://github.com/pgvector/pgvector/issues/326#issuecomment-2069362693 and https://github.com/pgvector/pgvector/issues/409#issuecomment-1895470718

image

Results show that on the same machine (Graviton3):
- index build time for half vectors using default inner product function is better than for float32 vectors only at high number of cores. Possibly due to less locks/memory accesses/disk IO 
- index build time for SVE inner product function for halfvectore is better than with default inner product function for halfvectors at any number of cores. Even at serial build. 

2. I tried to add SVE optimization for converting float32<->float16 but found no performance gain in 
`insert into emb_f16_3 (vector) select vector::halfvec(1536) from emb;`

Possibly because using SVE intrinsic makes sense when we deal with many float16 number altogether as it's in dot product calculation. So it's not in the patch.

I suggest that if the performance gain for (1) is good enough we could try to add some architecture checks into the patch. I'm not convinced that the way I did is optimal one. So if anyone more experiences in these checks add some improvements, I appreciate it very much. 

And also maybe it's worth adding SVE optimized cosine and l2 distance and testing these cases separately.


ankane:
Hi @pashkinelfe, thanks for the PR, and sorry for the delay. The build improvement looks nice. My main concern (besides maintenance of more distance functions) is accumulation happens at half precision, which will produce different results than the current code. I'd be curious to see how performance looks if it's changed to single precision.

pashkinelfe:
@ankane thanks! I'll try to ann-benchmark index quality soon

ankane:
Cleaning up stale PRs (but happy to revisit if there's a way to significantly improve performance without changing the accumulation precision).

PR #552, opened 2024-05-09
Make vector clone targets configurable

### LINKED ISSUE #551: Make function multiversioning configurable
### Current Behavior

As of `v0.7.0`, we can use `make OPTFLAGS=""` to disable native builds while still enjoying auto vectorization on select distance computations thanks to function multiversioning on Linux.

### Desired Behavior

Currently, the clone targets are hard coded to be `"default", "fma"`. It would be nice to be able to change the targets during build time. Because the targets are hard coded, we need to do something like the following

```sh
$ sed -E 's/target_clones\(([A-Za-z0-9 ,"]+)\)/target_clones(\1, "avx", "avx2", "avx512f")/' -i src/vector.c
$ make OPTFLAGS=""
```

which is a fragile approach. Instead, we would like to have the following

```sh
$ make OPTFLAGS="" CPPFLAGS='-D <SOME_VAR>=\'<SOME_VALUE>\''
```

_without_ changing anything in the checked-out source tree.

### Suggestions

The hard-coded target clones could be defined in a preprocessor directive, which could be overridden during build time.

A PR will follow for further discussions.

aytekinar:
Added a new preprocessor directive to make function multiversioning on distance computations configurable, i.e., during build time, the user can provide `CPPFLAGS` to override the default target clones.

Modified README.md to explain the behavior.

Closes #551.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1026,6 +1026,22 @@ To compile for portability, use:
 make OPTFLAGS=""
 ```
 
+### Optimized Versions of Certain Distance Functions
+
+By default, pgvector uses [function multiversioning][gcc:fmv] for certain distance
+computations to enable [automatic vectorization][wiki:av]. Currently, pgvector
+generates two clones for each of these functions: `default` and `fma`.
+
+To generate different clone targets for these functions (e.g., for `default`, `fma`,
+`avx`, `avx2`, and `avx512f`), use:
+
+```sh
+make OPTFLAGS="" CPPFLAGS='-D VECTOR_TARGET_CLONES_TARGETS=\'"default", "fma", "avx", "avx2", "avx512f"\''
+```
+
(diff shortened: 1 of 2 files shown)


tureba:
I'm trying to understand why this direction would be the way to go.

I agree the list of targets is currently conservative, and could be improved. But if there is a better list, then surely we would all like for it to be in pgvector itself, not in someone else's build scripts.

And if your intended user has deeper knowledge of better targets, we could just benchmark those and get them into pgvector.

So why propose to make it build-time configurable instead of suggesting an improved target list, one that you imagine your intended user would build pgvector?

ankane:
Hi @aytekinar, thanks for the PR. I agree with @tureba that it'd be better to benchmark and improve the target list if possible. If someone decides to do this, please include benchmarks for both index build time and query time comparing the existing and proposed targets in the PR.

PR #598, opened 2024-06-16
Remove redundant PGDLLEXPORT qualifiers.

higuoxing:
For functions being declared with PG_FUNCTION_INFO_V1(), the macro has added the PGDLLEXPORT qualifier automatically[^1].  We don't need to add them manually.

[^1]: https://github.com/postgres/postgres/commit/8cf64d35eaccab967c51d8bccf2c1babe3f84715


ankane:
Hi @higuoxing, thanks for the PR! Pushed a version of this in the commit above that also supports Postgres 12 - 15.

PR #604, opened 2024-06-23
Add direct array to/from sparsevec casts

Ngalstyan4:
Hi all,

We ran into an issue when trying to use the `sparsevec` type at 
Lantern with ~30K dimensional vectors (SPLADE) relevant commit message is below:

      Before this commit the only ways to construct a sparse vector was to
      cast one from vector/halfvec or create one from a literal.
      
      Neither approach is suitable for construcitng sparse vectors that have
      more than 16K dimensions, since intermediate ::vector cast disallows
      such large vectors and writing down such large vector in literal format
      is inconvenient.
      
      This is a needless limitation since both the target sparsevec type and
      the common source ARRAY type support larger dimensions.
      
      This commit overcomes the limitation by implementing direct
      ARRAY[]<->sparsevec casts

### Notes on the code

1. All current vector casts in pgvector disallow NULLs in ARRAYs that get cast to vectors. I followed that in the base commit. But also added a commit supporting NULL elements in array->sparsevec casts and treating NULLs as unset sparsevec elements
2. There is quite some repetition in `array_to_sparsevec` function when handling generic array to `float4` casts. There is similarly repeated code for [halfvec](https://github.com/pgvector/pgvector/blob/d1694a93afe4d23d0bc3ca6f9c85f06fb79a325c/src/halfvec.c#L479-L504) and [vector](https://github.com/pgvector/pgvector/blob/d1694a93afe4d23d0bc3ca6f9c85f06fb79a325c/src/vector.c#L482-L507) casts. I tried to reduce some of the repetition with the `PROCESS_ARRAY_ELEM ` macro
There is an alternative approach [here](https://github.com/pgvector/pgvector/commit/28282d0dcfdcaebe57a412477683c0606c25c813) where I factor out generic array element to float4 conversion to a function. The code is a lot cleaner there, but performance suffers. I benchmarked it with the query below:
`SELECT array_to_sparsevec(array_agg(0)::float4[], 16001000, false) FROM generate_series(1, 16001000) n;`
The version in this PR spends `60.0 +/- 1.2 ms` in the array extraction loop, while the linked version which uses a branch in each iteration of the loop takes `87.2 +/- 0.4 ms` on my 2019 Mac. Here is [the branch I used for the benchmarks](https://github.com/Ngalstyan4/pgvector/tree/array-to-sparsevec-benchmark).
I can probably combine something like the current `PROCESS_ARRAY_ELEM` with a macro-function similar to the static function from the benchmarking branch to reduce repetition with no performance hit. I am not sure the complexity is worth it
3. There is also a somewhat unrelated commit (8f2b7fec39a0ced45a76b97a932312dfceff3170) in the PR that changes one of the error messages to be closer to the error message [shown in the upstream](
https://github.com/search?q=repo%3Apostgres%2Fpostgres+%22array+must+be+one-dimensional%22&type=code) in similar situations

### NOT DONE:

1. [x] ~~Decide whether to support arrays containing NULLs to sparsevec casts (drop corresponding commit if not supporting)~~
1. [ ] Decide whether to simplify (at the cost of performance, or otherwise) the generic array to float4 conversion code
1. [x] ~~Decide whether sparsevec->array casts should contain zero values (current approach) or NULLs in place of `sparsevec`s empty slots in the returned ARRAY~~
1. [x] ~~Add update scripts introducing the functions and the casts~~

Happy to address any issues in this PR if you think it is useful to have it in pgvector


jkatz:
Going to address a few of the general notes:

>  Decide whether to support arrays containing NULLs to sparsevec casts (drop corresponding commit if not supporting)

By definition, a vector can't contain NULL attributes, so I don't think we should support this.

>  Decide whether to simplify (at the cost of performance, or otherwise) the generic array to float4 conversion code

I'm aligned with simplification only if it doesn't impact correctness or performance.

>  Decide whether sparsevec->array casts should contain zero values (current approach) or NULLs in place of sparsevecs empty slots in the returned ARRAY

I think it's on the user to decide how they want to treat this case - we can't assume that a `NULL` is meant to be `0`. A `NULL` means we don't know what value should be present.

Ngalstyan4:
Thanks for your notes, @jkatz !

I, for some reason, thought (wrongly) that postgres stores ARRAY NULLs in a bitmap, and so NULLs take less space than zeros, that is why proposed the approach.
Addressed the PR and made NULL handling identical to other vector types.

I also added an update script and fixed the build issue on all PGs.

(Will force-push and clean up git history when reviewing is done)

ankane:
Hi @Ngalstyan4, thanks for another PR! Will consider adding the array -> `sparsevec` casts in 0.8.0 (and give more feedback at that time). I'm hesitant to add a `sparsevec` -> `real[]` cast since it could create very large arrays (`SELECT '{}/134217727'::sparsevec::real[];`), but maybe that's not an issue.

ankane:
Hi @Ngalstyan4, merged the array to `sparsevec` casts in the commit above (and plan to release with 0.8.0 - #677). I'd like to hold off on `sparsevec` -> `real[]` (still have the same thoughts as the previous comment). Thanks for the PR!

PR #614, opened 2024-07-03
Add a compiler flag to use 512-bit vector instructions

rakshitgit-os:
This PR adds a compiler flag to use 512-bit instructions in auto-vectorized code (-mprefer-vector-width=512).

- Adding the flag is optional and controlled by a variable. To add the flag, set `USE_512BIT_VECTOR=1` when running make. By default, the flag is not added and the existing behavior is unchanged.
- Without this flag, we observe that several compilers (gcc11/12, clang13/14/15) use 256-bit registers even on newer platforms (such as AWS m7i instances). This is observed, for example, in distance functions.
- We observe performance benefits in certain conditions (platform/dataset) by adding this flag. For example, using ANN benchmarks with the gist-960 dataset, we observe 5% qps increase on m7i instances with the flag vs without.
- Note that `-mprefer-vector-width=512` is already used by default in [ANN benchmarks](https://github.com/erikbern/ann-benchmarks/blob/main/ann_benchmarks/algorithms/pgvector/Dockerfile) when building pgvector.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,11 @@ REGRESS_OPTS = --inputdir=test --load-extension=$(EXTENSION)
 
 # To compile for portability, run: make OPTFLAGS=""
 OPTFLAGS = -march=native
+ifeq ($(shell uname -m), x86_64)
+        ifeq ($(USE_512BIT_VECTORS), 1)
+            OPTFLAGS += -mprefer-vector-width=512
+        endif
+endif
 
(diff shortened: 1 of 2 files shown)


ankane:
Hi @rakshitgit-os, thanks for the PR. Users can use `OPTFLAGS="-march=native -mprefer-vector-width=512"`, so I don't think an additional flag is needed.

PR #639, opened 2024-08-08
Combine overflow checks and computation steps.

higuoxing:
Some of the overflow checks can be combined with the computation step and can save some loops.


ankane:
Hi @higuoxing, thanks for the PR. However, this breaks auto-vectorization (see the `Makefile` for flags to debug).

PR #193, opened 2023-07-22
Add Alpine Dockerfile

jramos:
This adds a Dockerfile for creating an Alpine build. The resulting image is ~80MB smaller than the Debian image.

PR #655, opened 2024-08-23
Docker image for Alpine (92.73 MB vs 147.31 MB)

SamuelMarks:
I've been [building](https://github.com/SamuelMarks/pgvector-builds) Alpine versions of your project https://hub.docker.com/r/samuelmarks/pgvector-builds/tags. It's 62% of [the size](https://hub.docker.com/r/pgvector/pgvector/tags); so IMHO probably good to include in your repo properly and release on your official hub.docker.com.


ankane:
Hi @SamuelMarks, thanks for another PR, but I still have the same thoughts as #193.

PR #658, opened 2024-08-23
Swap max costing values to align with upstream guidance

jkatz:
A feature targeted for PostgreSQL 18 (postgres/postgres@e2225346) that makes optimizations around disabled path nodes impacted pgvector such that PostgreSQL would choose to perform an index scan when it should have used a different scan (e.g. `SELECT count(*) FROM table`). Per upstream guidance[1], the recommendation is to switch to using `get_float8_infinity()`, which achieves the same behavior in backbranches, and can be adapated to work with the new behavior introduced in PostgreSQL 18.

[1] https://www.postgresql.org/message-id/2281822.1724441531%40sss.pgh.pa.us

### DIFF
--- a/src/hnsw.c
+++ b/src/hnsw.c
@@ -9,6 +9,7 @@
 #include "commands/vacuum.h"
 #include "hnsw.h"
 #include "miscadmin.h"
+#include "utils/float.h"
 #include "utils/guc.h"
 #include "utils/selfuncs.h"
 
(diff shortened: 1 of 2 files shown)


ankane:
Thanks @jkatz, will merge once things are finalized.

ankane:
Looks like CI is still failing for pg18, but should be okay to merge. Thanks @jkatz!

ankane:
Pushed a different fix in 1291b12090bbb03bd92b92e42a1567ae5b1c96ad to address this.

PR #679, opened 2024-09-23
Hnsw streamed tests

VladlenPopolitov:
Added tests fro hnsw streaming. Currently test crashes the server after TRUNCATE TABLE with active index and hnsw.streaming=ON.

### DIFF
--- a/test/expected/hnsw_streaming.out
+++ b/test/expected/hnsw_streaming.out
@@ -1,0 +1,100 @@
+SELECT t,
+Array[t,t+1,t+2]::vector(3) AS val
+INTO test_where
+FROM generate_series(1, 2000) as t;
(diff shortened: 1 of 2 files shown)


ankane:
Hi @VladlenPopolitov, thanks for the PR. I added basic tests for iterative search in b26a21b848fdb0a642fa16053ebdd1baddac0e25 (as the TAP tests should cover the main functionality), but it'd be good to include a test for `TRUNCATE` in `hnsw_vector.sql` as well.

VladlenPopolitov:
@ankane  I added other test #698, that checks the insertion of the first row to a table with index. It works in pair with TRUNCATE test.

ankane:
Added minimal tests for iterative index scans with empty indexes in the commit above.

PR #682, opened 2024-09-25
Update HNSW cost estimatation to utilize search and index info

jkatz:
Previously, the cost estimation formula for a HNSW index scan utilized
a methodology that only factored in the entry level for an HNSW scan
and the "m" index parameter, which reflects the number of tuples (or
vectors) to scan at each step of a HNSW graph traversal. While this
would bias the PostgreSQL query planner to choose an HNSW index scan
over other available paths, this could lead to potential suboptimal
index selection, for example, choosing to use a HNSW index instead of
an available B-tree index that has better selectivity.

The number of tuples scanned during HNSW graph traversal is principally
influenced by these factors:

 * The number of tuples stored in the index
 * `m` - the number of tuples that are scanned in each step of the graph
   traversal
 * `hnsw.ef_search` - which influences the total number of steps it
   takes for the scan to converge on the approximated nearest neighbors

Through testing different source models for vectors, we also observed
that the correlation of vectors in mdoels would impact this convergence.
For this first iteration, we've opted to hardcode a constant scaling
factor and set it to `0.55`, though a future commit may turn this into
a configurable parameter.

The high-level formula for estimating the cost of a HNSW index scan is
as such:

```
(entryLevel * m) + (layer0TuplesMax * layer0Selectivity)
```

where

- `(entryLevel * m)` is the lower bound of tuples to scan, as it
accounts for the graph traversal to layer 0 (L0). (L1 and above has an ef=1)
- `layer0TuplesMax` is an estimate of the maximum number of tuples to
scan at L0. This accounts for tuples that may end up being discarded due
to them already being visited. Testing shows that the number of steps
until converge is similar to the value of `hnsw.ef_search`, thus we can
estimate tuples max at `hnsw.ef_search * m * 2`
- `layer0Selectivity` - estimates the percentage of tuples that will
actually be scanned during the index traversal, multipled by the scaling
factor

In addition to the `m` build parameter and `hsnw.ef_search`, costs
estimates can be influenced by standard PostgreSQL costing parameters,
though adjusting those (e.g. `random_page_cost`) should be done with
care.

Co-authored-by: @ankane

### DIFF
--- a/src/hnsw.c
+++ b/src/hnsw.c
@@ -99,7 +99,10 @@ hnswcostestimate(PlannerInfo *root, IndexPath *path, double loop_count,
 {
 	GenericCosts costs;
 	int			m;
-	int			entryLevel;
+	int		entryLevel;
+	int		layer0TuplesMax;
+	double		layer0Selectivity;
+	double		scalingFactor = 0.55;
 	Relation	index;
 
 	/* Never use index without order */
@@ -119,12 +122,42 @@ hnswcostestimate(PlannerInfo *root, IndexPath *path, double loop_count,
 	HnswGetMetaPageInfo(index, &m, NULL);
 	index_close(index, NoLock);
 
-	/* Approximate entry level */
-	entryLevel = (int) -log(1.0 / path->indexinfo->tuples) * HnswGetMl(m);
+	/*
+	 * HNSW cost estimation follows a formula that accounts for the total
+	 * number of tuples indexed combined with the parameters that most influence
+	 * the duration of the index scan, namely:
+	 *   m - the number of tuples that are scanned in each step of the HNSW
+	 *       graph traversal
+	 *   ef_search - which influences the total number of steps taken at layer 0
+	 *
+	 *  The source of the vector data can impact how many steps it takes to
+	 *  converge on the set of vectors to return to the executor. Currently,
+	 *  we use a hardcoded scaling factor (HNSWScanScalingFactor) to help
+	 *  influence that, but this could later become a configurable parameter
+	 *  based on the cost estimations.
+	 *
+	 * The tuple estimator formula is below:
+	 *
+	 * numIndexTuples = (entryLevel * m) +
+	 * 					(layer0TuplesMax * layer0Selectivity)
+	 *
+	 * "entryLevel * m" represents the floor of tuples we need to scan to get
+	 * to layer 0 (L0).
+	 *
+	 * "layer0TuplesMax" is the estimated total number of tuples we'd scan at
+	 * L0 if we weren't discarding already visited tuples as part of the scan.
+	 *
+	 * "layer0Selectivity" estimates the percentage of tuples that are scanned
+	 * at L0, accounting for previously visited tuples, multiplied by the
+	 * "scalingFactor" (currently hardcoded).
+	 */
+	entryLevel = (int) floor(log(path->indexinfo->tuples + 1) * HnswGetMl(m));
+	layer0TuplesMax = HnswGetLayerM(m, 0) * hnsw_ef_search;
+	layer0Selectivity =  (scalingFactor * log(path->indexinfo->tuples + 1)) /
+		(log(m) * (1 + log(hnsw_ef_search)));
 
-	/* TODO Improve estimate of visited tuples (currently underestimates) */
-	/* Account for number of tuples (or entry level), m, and ef_search */
-	costs.numIndexTuples = (entryLevel + 2) * m;
+	costs.numIndexTuples = (entryLevel * m) +
+		(layer0TuplesMax * layer0Selectivity);
 
 	genericcostestimate(root, path, loop_count, &costs);
 
--- a/test/t/017_hnsw_filtering.pl
+++ b/test/t/017_hnsw_filtering.pl
@@ -117,8 +117,8 @@ $node->safe_psql("postgres", "CREATE INDEX attribute_idx ON tst (c);");
 $explain = $node->safe_psql("postgres", qq(
 	EXPLAIN ANALYZE SELECT i FROM tst WHERE c = $c ORDER BY v <-> '$query' LIMIT $limit;
 ));
-# TODO Use attribute index
-like($explain, qr/Index Scan using idx/);
+# Use attribute index
+like($explain, qr/Bitmap Index Scan on attribute_idx/);
 
 # Test partial index
 $node->safe_psql("postgres", "CREATE INDEX partial_idx ON tst USING hnsw (v vector_l2_ops) WHERE (c = $c);");


jkatz:
Note: I've fixed a previous test failure (it was a TODO that's now resolved with the updated cost estimation model :smile:). However, the active failure is interesting. The failing query is:

```sql
EXPLAIN ANALYZE SELECT i FROM tst WHERE v <-> '$query' < 1 ORDER BY v <-> '$query';
```

The dataset size is 10K vectors, and to explicitly state, this is a query with a filter that is ordering the vectors but doesn't have an explicit limit. `hnsw.ef_search` is `40`.

Previously, the costing was such that PostgreSQL estimated the sequential scan was cheaper and opted for that.
With the current patch, PostgreSQL estimates the index scan is cheaper (by almost less than a single cost unit 😨) and opts to use that. This may very well be the fastest scan path, but (untested theory) in streaming mode would this end up using the full index scan, which would certainly be more expensive than a sequential scan?

I have a locally modified version of the patch that borrows similar selectivity logic to IVFFlat (`MIN(our estimate, PostgreSQL's selectivity estimate)`) but PostgreSQL still opted for the sequential scan.

Primarily posting this as a FYI. I'm still pulling a few threads.

src/hnsw.c
```diff
@@ -119,15 +121,53 @@ hnswcostestimate(PlannerInfo *root, IndexPath *path, double loop_count,
...
 
 	genericcostestimate(root, path, loop_count, &costs);
 
+	/*
+	 * Explicitly set overall index selectivity after populating cost estimates,
+	 * as the generic cost estimator only factors in predicates as part of the
+	 * elimination strategy, whereas HNSW selecitivty is based on scanning the
+	 * index itself.
+	 */
+	costs.indexSelectivity = costs.numIndexTuples / path->indexinfo->tuples;
```
ankane, line 169: Removing this should fix the test. I don't think this type of selectivity is what the planner expects.

I think we should do the same for IVFFlat ([ivfflat-cost branch](https://github.com/pgvector/pgvector/compare/ivfflat-cost)), which fixes two of the tests.

src/hnsw.c
```diff
@@ -119,15 +121,53 @@ hnswcostestimate(PlannerInfo *root, IndexPath *path, double loop_count,
...
+	 *
+	 * "layer0TuplesMax" is the estimated total number of tuples we'd scan at
+	 * L0 if we weren't discarding already visited tuples as part of the scan.
+	 *
+	 * "layer0Selectivity" estimates the percentage of tuples that are scanned
+	 * at L0, accounting for previously visited tuples.
+	 *
+	 * "HnswSF" is a scaling factor, hardcoded to HnswScanScalingFactor
+	 */
+	totalLayers = floor(log(path->indexinfo->tuples) / log(m));
```
ankane, line 154: I think it'd be better to keep the previous naming (`entryLevel`) here, as well as `HnswGetMl(m)` since it's really the total number of layers - 1.

We may also want to add `+ 1` to the number of tuples here and below to avoid `log(0)`.

src/hnsw.c
```diff
@@ -119,15 +121,53 @@ hnswcostestimate(PlannerInfo *root, IndexPath *path, double loop_count,
...
+	 *  The source of the vector data can impact how many steps it takes to
+	 *  converge on the set of vectors to return to the executor. Currently,
+	 *  we use a hardcoded scaling factor (HNSWScanScalingFactor) to help
+	 *  influence that, but this could later become a configurable parameter
+	 *  based on the cost estimations.
+	 *
+	 * The tuple estimator formula is below:
+	 *
+	 * numIndexTuples = (totalLayers * m) +
+	 * 					(layer0TuplesMax * layer0Selectivity * HnswSF)
```
ankane, line 141: I think `HnswSF` should be included in `layer0Selectivity` to make the `layer0Selectivity` description accurate.

src/hnsw.h
```diff
@@ -85,6 +85,9 @@
 /* Optimal ML from paper */
 #define HnswGetMl(m) (1 / log(m))
 
+/* Scaling factor for HNSW index scan estimation */
+#define HnswScanScalingFactor ((double) 0.55)
```
ankane, line 89: I'd prefer to keep this in `hnswcostestimate`, but if you feel strongly it should be separate, let's use the prefix `HNSW_COST`.

ankane (commented):
Awesome, thanks for driving this @jkatz! This should significantly improve index selection.

Added some comments inline.

src/hnsw.c
jkatz: @ankane I'm less opinionated on the name; I did try to force `tuples` to 0, but in that case PostgreSQL won't consider the scan. I think in the case where `tuples` is 0, we should max out the costs like we do if there's no ORDER BY quals.

src/hnsw.c
jkatz: Going to try that; if it works will push up with next commit.

src/hnsw.h
jkatz: I don't feel strongly about that - but - I do feel it should be configurable and just wanted to simplify transition if/when it happens 😛 I will move.

jkatz:
@ankane Pushed changes inline with feedback. One discussion point will be the check to push the costs up to the max if tuples is 0.

I'm...OK-ish to go with your suggestion (add `1` to the total number), but arguably if we have no tuples, we shouldn't even bother considering the index scan. Adding 1 is certainly a cheap operation, but it'd be cheaper in that case to not even go through all the operations. OTOH, we then have to check that condition every time, and given it's unlikely we'd encounter that case, it may just be better to "add 1".

src/hnsw.c
```diff
@@ -99,11 +99,18 @@ hnswcostestimate(PlannerInfo *root, IndexPath *path, double loop_count,
...
+	double		layer0TuplesMax;
+	double		layer0Selectivity;
+	double		hnswScalingFactor = 0.55;
 	Relation	index;
 
-	/* Never use index without order */
-	if (path->indexorderbys == NULL)
+
+/* Scaling factor for HNSW index scan estimation */
+#define HnswScanScalingFactor ((double) 0.55)
```
ankane, line 110: Think this was left over.

src/hnsw.c
```diff
@@ -99,11 +99,18 @@ hnswcostestimate(PlannerInfo *root, IndexPath *path, double loop_count,
 {
 	GenericCosts costs;
 	int			m;
-	int			entryLevel;
+	double		entryLevel;
+	double		layer0TuplesMax;
+	double		layer0Selectivity;
+	double		hnswScalingFactor = 0.55;
```
ankane, line 105: Let's remove the `hnsw` prefix here and in the comments for simplicity.

src/hnsw.c
```diff
@@ -119,12 +126,42 @@ hnswcostestimate(PlannerInfo *root, IndexPath *path, double loop_count,
...
+	 * to layer 0 (L0).
+	 *
+	 * "layer0TuplesMax" is the estimated total number of tuples we'd scan at
+	 * L0 if we weren't discarding already visited tuples as part of the scan.
+	 *
+	 * "layer0Selectivity" estimates the percentage of tuples that are scanned
+	 * at L0, accounting for previously visited tuples, multiplied by the
+	 * "hnswScalingFactor" (currently hardcoded).
+	 */
+	entryLevel = floor(log(path->indexinfo->tuples) / log(m));
```
ankane, line 158: Let's use `* HnswGetMl(m)` instead of `/ log(m)` like the previous code. `ml` will likely never change, but it's really a function of that.

ankane (commented):
Great, added a few more comments. I think the `+ 1` approach would be slightly better since the index can fulfill the query (the first path is more designed to prevent something the index can't do).

src/hnsw.c
jkatz: @ankane Any reason outside of matching the paper? This way saves a `/` operation.

ankane: It only saves a `1 *` which the compiler likely removes, but I think it's more correct. If `ml` changes, we would need to change this.

jkatz: @ankane Alright, I concede :smile: Pushing it up in a sec.

src/hnsw.c
jkatz: 🤦

jkatz:
@ankane All feedback should now be addressed.

ankane:
Thanks, looks great. Last remaining comment:

> I think the + 1 approach would be slightly better since the index can fulfill the query (the first path is more designed to prevent something the index can't do).

jkatz:
@ankane Done.

src/hnsw.c
```diff
@@ -99,10 +99,13 @@ hnswcostestimate(PlannerInfo *root, IndexPath *path, double loop_count,
...
 	int			m;
-	int			entryLevel;
+	double		entryLevel;
+	double		layer0TuplesMax;
+	double		layer0Selectivity;
+	double		scalingFactor = 0.55;
 	Relation	index;
 
-	/* Never use index without order */
+	/* Never use index without order or if there are no tuples */
```
ankane, line 108: This comment needs reverted.

jkatz: 🤦 Sorry for the sloppiness - that's due to rushing. Pushed fix.

src/hnsw.c
```diff
@@ -119,12 +122,42 @@ hnswcostestimate(PlannerInfo *root, IndexPath *path, double loop_count,
...
+	 * to layer 0 (L0).
+	 *
+	 * "layer0TuplesMax" is the estimated total number of tuples we'd scan at
+	 * L0 if we weren't discarding already visited tuples as part of the scan.
+	 *
+	 * "layer0Selectivity" estimates the percentage of tuples that are scanned
+	 * at L0, accounting for previously visited tuples, multiplied by the
+	 * "scalingFactor" (currently hardcoded).
+	 */
+	entryLevel = floor(log(path->indexinfo->tuples) * HnswGetMl(m));
```
ankane, line 154: Looks like `+ 1` is missing from here.

src/hnsw.c
```diff
@@ -99,7 +99,10 @@ hnswcostestimate(PlannerInfo *root, IndexPath *path, double loop_count,
 {
 	GenericCosts costs;
 	int			m;
-	int			entryLevel;
+	double		entryLevel;
+	double		layer0TuplesMax;
```
ankane, line 103: I think the code will be cleaner if an `int` is used for `entryLevel` and `layer0TuplesMax`, but I can do that in a follow-up commit if that's easier.

src/hnsw.c
```diff
@@ -119,12 +122,42 @@ hnswcostestimate(PlannerInfo *root, IndexPath *path, double loop_count,
...
+	 * L0 if we weren't discarding already visited tuples as part of the scan.
+	 *
+	 * "layer0Selectivity" estimates the percentage of tuples that are scanned
+	 * at L0, accounting for previously visited tuples, multiplied by the
+	 * "scalingFactor" (currently hardcoded).
+	 */
+	entryLevel = floor(log(path->indexinfo->tuples) * HnswGetMl(m));
+	layer0TuplesMax = (double) (HnswGetLayerM(m, 0) * hnsw_ef_search);
+	layer0Selectivity =  scalingFactor * 
+		(log(path->indexinfo->tuples + 1) / (log(m) * (1 + log(hnsw_ef_search))));
```
ankane, line 157: Just noticed there's an extra space here and we can get rid of some parentheses here and below.

ankane (commented):
Added some more comments, but I can handle the style-related ones in a follow-up commit if that's easier.

jkatz:
@ankane Addressed the feedback. I adjusted the parens, but ensuring it aligns to the formula.

image

ankane:
Awesome, thank you! And huge thanks for all of the work that went into deriving the formula!!

jkatz:
@ankane 🎉Thanks! Looking forward to describing more of the data...and adding the scaling factor parameter 😉

PR #698, opened 2024-10-23
HNSW index test for insert to empty table with empty index is added

VladlenPopolitov:
As soon as TRUNCATE TABLE test is added to check creation of index on empty table, it would be good to add the test of inserting first row to empty table with created index (no one other tests cover this situation).


ankane:
Hi @VladlenPopolitov, thanks for the PR, but there are TAP tests that cover this.

PR #755, opened 2025-01-10
chore(ci): update to use pg 17

chenrui333:
- bump to use pg 17
- update macos-14 with macos-15 (both are arm runners)


ankane:
Thanks @chenrui333!

PR #764, opened 2025-01-24
Avoid -march=native when reproducible build is wanted

bmwiedemann:
With this patch, compilation will avoid `-march=native` when a reproducible build is wanted.

See https://reproducible-builds.org/ for why this is good.

Without this patch, compiling on different machines produced different binaries, which made verification of results difficult.

This patch was done while working on reproducible builds for openSUSE.


bmwiedemann:
Yes, I saw that in the README, but then every distribution has to discover and apply that (we did not). With this patch on the other hand, it would already be the default for us.

Issue #745, opened 2024-12-31
Automate estimation of amount of memory needed for fast HNSW index build


michaelbautin:
It would be very useful to have a function that would estimate, for a particular combination of vector data type, dimensionality, number of rows, and the value of M, the minimum amount of memory (as represented by maintenance_work_mem) necessary to build the HNSW index fully in memory. In many cases, users would like to avoid the case of spilling the index to disk at build time altogether, because that often slows down the index build too much.

jkatz:
Thanks for the suggestion.

The rough math I typically use is `# vectors * sizeof(VECTOR) * 1.3`, where `VECTOR` is the data type size (e.g. `vector(1024)` would be `1024*4 + 8`).

So for example, if I have 1,000,000 1024-dim fp32 vectors, the math would be:

```
1000000 * (1024 * 4 + 8) * 1.3 => 5,335,200,000
```

which to get into GiB would be:

```
5,335,200,000 / 1024**3 => ~4.96GB
```

This can get more nuanced due to page alignment (i.e. how many vectors can fit within 8KB), how many layers you have (which is dictated by `m`), and quantization, but I at least wanted to give you a simple formula to start with.

michaelbautin:
Would this estimation differ at all for parallel vs. non-parallel build? From looking at the code it seems to be that it should not.

jkatz:
Nope, it's the same regardless of workers.

ankane:
Hi @michaelbautin, also check out #299.

michaelbautin:
Thanks! In #299, the function  being linked to is 
```c
/*
 * Get the max number of elements that fit into maintenance_work_mem
 */
static double
HnswGetMaxInMemoryElements(int m, double ml, int dimensions)
{
	Size		elementSize = sizeof(HnswElementData);
	double		avgLevel = -log(0.5) * ml;

	elementSize += sizeof(HnswNeighborArray) * (avgLevel + 1);
	elementSize += sizeof(HnswCandidate) * (m * (avgLevel + 2));
	elementSize += sizeof(ItemPointerData);
	elementSize += VECTOR_SIZE(dimensions);
	return (maintenance_work_mem * 1024L) / elementSize;
}
```

I see that that function is not present in the current version (v0.8.0) and the actual amount of memory consumed is being used for deciding when to start spilling to disk instead. I think it would be good to have a reliable function, accessible to the user, that could be used to estimate whether a particular HNSW index build could be completed in memory.

[...]

Issue #500, opened 2024-03-22
Index Build Time Does Not Improve as Expected When Changing "Workers"


iamsabhoho:
Hello, 

I'm Sabrina, a data scientist at GSI Technology. I've been working on a pgvector benchmarking project on deep-1B dataset. 

The documentation states that I can improve index build time by changing the parameters **max_parallel_maintenance_workers** and **max_parallel_workers**. 

Unfortunately, if I increase the values of the parameters beyond 10, I do not see an improvement. We tested with 1M records of deep-1B and here are some results: 
| number of workers | index build time in seconds |
| ----------- | ----------- |
| default | 4041 |
| 10 | 2043 |
| 20 | 2090 |

I am using the following hardware:
- Intel(R) Xeon(R) Gold
- 104 cores

Here are the loaded extensions:
| Name   | Version |   Schema   |                     Description                   |   
| ----------- | ----------- |  ----------- | ----------- |
| plpgsql | 1.0     | pg_catalog | PL/pgSQL procedural language | 
| vector  | 0.6.1   | public     | vector data type and ivfflat and hnsw access methods |

Should I expect to see better performance as I increase the value of workers? Thank you!

I am happy to provide my code or any additional machine information needed.

ankane:
Hi @iamsabhoho, pgvector 0.6.2 reduces lock contention for HNSW builds with a large number of parallel workers, so I'd recommend trying that. Also, the build times look very slow compared to previous benchmarks (https://github.com/pgvector/pgvector/issues/409), so I'd double check that `maintenance_work_mem` ([docs](https://github.com/pgvector/pgvector?tab=readme-ov-file#index-build-time)) and other settings like `shared_buffers` ([docs](https://github.com/pgvector/pgvector?tab=readme-ov-file#tuning)) are set correctly. With only 1M records, it's possible you won't see much difference with more than 10 workers.

iamsabhoho:
Hi @ankane, 
Thanks for the fast response! I couldn't find `shared_buffers` with the link provided. Could you recommend a setting for me to try it out? Thanks again!

ankane:
`shared_buffers` should typically be 25% of your memory.

iamsabhoho:
@ankane thanks for the tips! We re-ran the experiments based on your suggestions. The build time matches your 1M benchmark closely. As you mentioned, adding more workers does not improve the time linearly. 

[...]

PR #765, opened 2025-01-24
Functions to estimate the amount of memory for building an HNSW index

michaelbautin:
This will allow users to verify if they can build an HNSW index fully in memory, which is frequently necessary to avoid unacceptable performance degradation, and to compute the appropriate value of maintenance_work_mem.

I am planning to expand this pull request with validation that the estimated memory usage is indeed correct during index build.

Fixes: #745

Related issues: #500

### DIFF
--- a/sql/vector.sql
+++ b/sql/vector.sql
@@ -916,3 +916,11 @@ CREATE OPERATOR CLASS sparsevec_l1_ops
 	OPERATOR 1 <+> (sparsevec, sparsevec) FOR ORDER BY float_ops,
 	FUNCTION 1 l1_distance(sparsevec, sparsevec),
 	FUNCTION 3 hnsw_sparsevec_support(internal);
+
+CREATE FUNCTION hnsw_build_memory_bytes_per_vector(integer, integer)
+	RETURNS float8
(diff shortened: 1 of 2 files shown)


ankane:
Hi @michaelbautin, thanks for the PR. If someone comes up with an approximate formula based on actual data from different datasets and index options, I may be open to including it in the readme, but don't want to add a function for it.

michaelbautin:
@ankane The formulas in this PR are very precise (less than 0.01% error in my experiments), as long as we know the vector sizes. For sparse vectors, where the vector size might vary, this would be more complicated. But to compute the same formulas outside of the pgvector code, we would have to know various sizeof values of internal data structures. So any memory sizing automation code would have to know these internal details.

michaelbautin:
@ankane 
>  If someone comes up with an approximate formula based on actual data from different datasets and index options
I am not sure what you mean by "formula based on actual data". The only parameters influencing the build HNSW in-memory size are M (given that mL is computed based on M), the number of vectors, and per-vector memory usage, assuming it is the same for all vector, which covers a lot of use cases. Given those parameters, the memory usage can be estimated with a very high precision and there is no need for any additional data collection from different datsets.

PR #772, opened 2025-02-03
Optimise vector distance functions for IBM Power Architecture

mikejuliet13:
### Summary

This PR optimises vector distance functions in PGVector, specifically:
- Cosine Similarity
- Inner Product
- L2 Squared Distance

By leveraging SIMD (Single Instruction, Multiple Data), these functions have been vectorizsd, reducing computation time and improving performance.

### Implementation Details:

- Used Altivec/VSX for IBM Power architecture.
- Applied vector-based optimizations to process multiple elements simultaneously.
- Reduced redundant computations, leading to overall efficiency gains.

### Testing & Benchmarking

Functionality Testing:
Verified correctness on both Power and x86 architectures.
Performance Testing:
Benchmarked on IBM Power using ANN-Benchmark, showing significant improvements in execution time.

### Impact
This optimisation enhances the efficiency of PGVector’s vector search operations, making it faster and more scalable for workloads requiring high-performance similarity searches.

### DIFF
--- a/src/vector.c
+++ b/src/vector.c
@@ -35,6 +35,11 @@
 #define VECTOR_TARGET_CLONES
 #endif
 
+#if defined(__powerpc__)
+#include <altivec.h>   /* Required for the Power GCC built-ins  */
+#define FLOAT_VEC_SIZE 4
+#endif
+
 PG_MODULE_MAGIC;
 
 /*
@@ -542,6 +547,43 @@ halfvec_to_vector(PG_FUNCTION_ARGS)
 	PG_RETURN_POINTER(result);
 }
 
+#ifdef __powerpc__
+
+VECTOR_TARGET_CLONES static float VectorL2SquaredDistance(int dim, float *ax,
+                                                          float *bx) {
+  size_t i;
+
+  float res = 0;
+  /* Vector implmentaion uses vector size of FLOAT_VEC_SIZE.  If the input
+         array size is not a power of FLOAT_VEC_SIZE, do the remaining elements
+         in scalar mode.  */
+  size_t base;
+
+  vector float *vx, *vy;
+  vector float vtmp = {0, 0, 0, 0};
+  vector float vres = {0, 0, 0, 0};
+
+  base = (dim / FLOAT_VEC_SIZE) * FLOAT_VEC_SIZE;
+
+  for (i = 0; i < base; i = i + FLOAT_VEC_SIZE) {
+    vx = (vector float *)(&ax[i]);
+    vy = (vector float *)(&bx[i]);
+
+    vtmp = vx[0] - vy[0];
+    vres += vtmp * vtmp;
+  }
+
+  /* Handle any remaining data elements */
+  for (i = base; i < dim; i++) {
+    const float tmp = ax[i] - bx[i];
+    res += tmp * tmp;
+  }
+
+  return res + vres[0] + vres[1] + vres[2] + vres[3];
+}
+
+#else
+
 VECTOR_TARGET_CLONES static float
 VectorL2SquaredDistance(int dim, float *ax, float *bx)
 {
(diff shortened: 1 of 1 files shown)


jkatz:
> Benchmarked on IBM Power using ANN-Benchmark, showing significant improvements in execution time.

Can you show the output of these tests, with the before and after?

mikejuliet13:
> > Benchmarked on IBM Power using ANN-Benchmark, showing significant improvements in execution time.
> 
> Can you show the output of these tests, with the before and after?

From ann-benchmark, after setting the metric for Cosine and Euclidean (L2), I have captured the data in pictorial form for before and after the execution of testcases.

For Codine+InnerProduct metric:
comparison_graph (1)

For L2 Squared Distance metric:
pg_l2_comparison_graph (1)

Here OG refers to the original code and Modified file refers to the vector optimised code

jkatz:
@mikejuliet13 Which dataset was this test run on? Was there any impact on recall?  What hardware was this on? What were the PostgreSQL settings used? Did you test this with any concurrency What were the results from other datasets?

Here's [an example](https://jkatz05.com/post/postgres/pgvector-performance-150x-speedup/) for how to report speedup results.

Also, is there a reason that POWER requires the explicit SIMD instructions that aren't available from the autovectorization gcc flags?

ankane:
Hi @mikejuliet13, thanks for the PR. As @jkatz mentioned, I think it'd be better to try [auto-vectorization](https://gcc.gnu.org/projects/tree-ssa/vectorization.html#using):

```sh
make OPTFLAGS="-maltivec"
```

If that works, we could probably add it [to the Makefile](https://github.com/pgvector/pgvector/blob/v0.8.0/Makefile#L27).

ankane:
Although looking at #117, this may not be needed.

ankane:
If there is a difference between this approach and auto-vectorization, I think it'd be better to submit the improvements to the compiler. Closing this out, but feel free to share any numbers if you have them.

mikejuliet13:
Hi @ankane  ,
Thanks for the follow-up.

My changes leverage Power-specific SIMD intrinsics (Altivec/VSX) along with manual loop unrolling to better utilize the architecture’s capabilities. While the -O3 flag in the original code enables auto-vectorization, it does not perform loop unrolling. This leads to a performance bottleneck—specifically around the xvmaddmsp instruction, which accounts for ~54% of the hotspots in perf, as operations repeatedly funnel through multiply-adds on the same register.

There is clear scope for performance improvement on Power by applying loop unrolling, either manually or via the -funroll-loops compiler flag.

Key Performance Observations (Power10):
Original code + -funroll-loops performs better than without it:
→ QPS improvement: 6–12%
→ Indexing time reduction: 10–20%
(observed primarily on higher-dimensional datasets like fashion-mnist-784-euclidean and gist-960-euclidean)

For lower-dimensional datasets like nytimes-256-angular and glove-100-angular, the performance difference is negligible for both QPS and indexing time.

Optimized code (manual VSX + manual loop unrolling) outperforms even the loop-unrolled original:
→ QPS improvement: 4–10%
→ Indexing time reduction: 10–30%
(notable gains on higher-dimensional datasets and even nytimes-256-angular)

For glove-100-angular, the difference remains minimal.

I tested across the following datasets in both batch and non-batch modes:
fashion-mnist-784-euclidean, gist-960-euclidean, nytimes-256-angular, and glove-100-angular.

Additional Insights:
Higher dimensionality correlates with greater performance benefits from the optimized code.

Larger m and ef_search values also amplify the performance gains in favor of the optimized version.

While auto-vectorization is certainly beneficial, these findings suggest that manual control—particularly with loop unrolling—still delivers tangible performance improvements on Power architecture.

I'm happy to share the full perf data or benchmarking logs if needed.

Issue #504, opened 2024-03-26
HNSW index building slow for large table


mavalderrama:
Hello, I've been working with pgvector for a while, and now we are giving a try to HNSW `0.6.0` latest version available on RDS, for a table with more than 170M rows and an embedding vector of 64 dimensions.

Our test bench:
- AWS instance `db.r6i.2xlarge`
- 64GB RAM
- 8 vcpu

We've made several attempts without success:

Default settings:
- Inserting data with the index in place: After 6 days, the progress was at 20%. I stopped the process
- Building the index with the data in place: After 4 days the progress was only ~24%. I stopped the process.

Setting `ef_construction` from `64` to `32`
- We've noticed the process was faster but after 4 days the progress was only at ~50%

Increasing workers:
- I didn't notice any real speed increase

What settings do you recommend guys?

Thanks in advance.

jkatz:
A few things:

* Be sure to increase `maintenance_work_mem` to ensure the build is occurring in memory.
* With 8 vCPU, you will want to set `max_parallel_maintenance_workers` to `7` (as by default you'll get 8 with this setting)
* https://github.com/pgvector/pgvector/commit/3ea2ce89be54d7060684be2c93501515f647a7ba which is available in 0.6.2 improves locking contention, which will speed up the builds.

This in turn should let you _increase_ `ef_construction` as you'll see performance improve.

ankane:
Hi @mavalderrama, also check out #500.

Issue #300, opened 2023-10-12
HNSW | Slow Build Times For DB With 10m Vectors


jordanparker6:
I have a HNSW index build that is taking +24hrs for a DB with 10M vectors of shape 768. 

The DB is running in GCP with pgvector 0.5.0 on a 4 CPU and 16 GB instance. CPU utilisation is hovering around 10%.

```
SET max_parallel_maintenance_workers = '8';
SET maintenance_work_mem = '1GB';
CREATE INDEX "documentChunk_hnsw_index" ON "DocumentChunk" 
USING hnsw (embedding vector_ip_ops);
```

Do you have any suggestions on how to speed this up and use more cores? 

Would the only resolutions be to change `m` and `ef_construction`? 

Also, to get the most of pgvector at query time, i need to set the `shared_buffers` to be greater than the hnsw index size and then `prewarm`? I think in my case that would be a 7GB * 10 = 70GB of RAM.

Happy to contribute to the README.md on practical performance configurations as I think it would be useful.

ankane:
Hi @jordanparker6, I'd recommend setting `maintenance_work_mem` much higher for the session (you'll see a notice when the index build no longer fits into memory and becomes slow). You could also create the index when the table is empty and insert the data in-parallel (w/ custom scripts).

Also, note that 0.5.1 speeds up HNSW builds significantly, and parallel index builds are planned for the future.

For queries, having enough memory for the entire index (and more) should be best for performance, and I'd recommend starting with general advice for shared buffers (25% of total memory). The index should warm up as it's used, but you can use pg_prewarm to eager load it.

jimmythepad:
Hey @jordanparker6 and @ankane, wondering if you gained any other insights on this?

We are facing a similarly long build time (24hrs) with a smaller vector and size (3.8m vectors with dimension 128) on GCP:

8cpu 52gb
postgres 14
using the built in cloudsql - 0.6.0 i think

(m = 100, ef_construction = 1000);

SET max_parallel_maintenance_workers = 7; -- plus leader
SET maintenance_work_mem = '8GB';

jkatz:
0.6.2 included an improvement to locking that will speed up your build.

PR #775, opened 2025-02-12
Change NOTICE to WARNING when HNSW graph does not fit in memory during build

michaelbautin:
This should help prevent cases where users wait for a very long time (24+ hours) for index build because they have not configured the maintenance work memory correctly, as happened in #300 and #504.


ankane:
Hi @michaelbautin, thanks for the PR, but don't see a strong reason to change this, as `NOTICE` is the default level for `client_min_messages`.

PR #782, opened 2025-02-22
chore(git): Update git ignore with .deps

kk-src:


ankane:
Hi @kk-src, I don't think this is needed, but please include a description in PRs.

kk-src:
After a make is done, .dep file starts to show up in git status.

```bash
$git status
On branch master
Your branch is up to date with 'origin/master'.

Untracked files:
  (use "git add <file>..." to include in what will be committed)
        .deps/

nothing added to commit but untracked files present (use "git add" to track)
```

When using vscode the individual files with .Po extension from .deps also show up with ~19 changes (as can be expected with no .gitignore of .deps).

@ankane : Can you please reconsider this with the above notes  :-) ?

ankane:
@kk-src I'm not sure how to reproduce (the installation instructions don't seem to generate a `.deps` directory). If it's editor-specific, consider adding it to your global `.gitignore`.

PR #783, opened 2025-02-22
chore(doc): Align README code block highlight with other sql markers

kk-src:


ankane:
Hi @kk-src, some of the code blocks use `tsql` for syntax highlighting since they aren't correctly highlighted with `sql`.

PR #802, opened 2025-03-18
Pgvector crossproduct

matteodurighetto:
Hello, this is a simple feature: cross product for 3 dimensional vectors, but it is useful for some physics and math calculations .

I have add this function : vector_cross_product (vector,vector) → vector | Cross Product right-handed | 0.8.0  
with the operator symbol <*>

It check if the vectors are the same dimensions and the dimension is 3, the calculate the new vector, then check it there are some overflows and if it's all correct, it returns the result vector. 

I have done some testing and it seems to work correctly

### DIFF
--- a/README.md
+++ b/README.md
@@ -963,6 +963,7 @@ Operator | Description | Added
 <#> | negative inner product |
 <=> | cosine distance |
 <+> | taxicab distance | 0.7.0
+<*> | cross product | 0.8.0
 
 ### Vector Functions
 
@@ -977,6 +978,7 @@ l2_normalize(vector) → vector | Normalize with Euclidean norm | 0.7.0
 subvector(vector, integer, integer) → vector | subvector | 0.7.0
 vector_dims(vector) → integer | number of dimensions |
 vector_norm(vector) → double precision | Euclidean norm |
+vector_cross_product(vector,vector) → vector | Cross Product right-handed | 0.8.0
 
 ### Vector Aggregate Functions
 
--- a/sql/vector.sql
+++ b/sql/vector.sql
@@ -111,6 +111,9 @@ CREATE FUNCTION vector_avg(double precision[]) RETURNS vector
 CREATE FUNCTION vector_combine(double precision[], double precision[]) RETURNS double precision[]
 	AS 'MODULE_PATHNAME' LANGUAGE C IMMUTABLE STRICT PARALLEL SAFE;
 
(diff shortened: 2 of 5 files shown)


jkatz:
From 5 minutes of search, I couldn't find any other PostgreSQL extensions that expose this (PostGIS has an [internal function](https://postgis.net/docs/doxygen/3.5/d2/ddd/lwgeodetic_8c_ab49bfa54fdded3ab7135b890d977cebf.html]) - primarily I'm trying to see if anyone has defined an operator for this.

Given the mathematical notation is `x` for cross product, I'd be hesitant to use `<*>` here (and I was also trying to find some prior art if PostgreSQL extensions used `<*>`; it's hard to search for operators 😢 ). However - while there's still discussion on the patch itself, I'd suggest we try to use an operator closer to the mathematical one, or a commonly accepted one, maybe `<x>`?

jkatz:
I may have bikeshedded on operator definitions too soon - I forgot the cross product was limited to 3 dimensions. I don't know if it makes sense to support this in pgvector, given pgvector supports n-dimensional vectors. If you need speed of calculation, you could use PL/Rust to define a cross-product function.

ankane:
Hi @matteodurighetto, thanks for the PR, but I don't think it's common enough to add. You can create a SQL function for it without pgvector.

```sql
CREATE FUNCTION cross_product(a real[], b real[]) RETURNS real[] AS $$
    SELECT ARRAY[a[2] * b[3] - a[3] * b[2], a[3] * b[1] - a[1] * b[3], a[1] * b[2] - a[2] * b[1]];
$$ LANGUAGE SQL;

SELECT cross_product('{1,0,0}', '{0,1,0}');
SELECT cross_product('{1,0,0}', '{0,0,1}');
SELECT cross_product('{3,-3,1}', '{4,9,2}');
```

matteodurighetto:
Hello @ankane @jkatz thank you to  have review this pull request/enhancement.  
I know it's possible to defines yourself the function, but I think it's quite useful to have this operator just available in some cases with the check of dimensions and so on, especially for some math/physical applications and some checks ( like to verify if the vectors are parallel  and you  check the result is 0 vector ).
Moreover it's a basic operation between vectors like inner product (that it's a function of pg_vector). 

Kind Regards

jkatz:
@matteodurighetto You can [define an operator](https://www.postgresql.org/docs/current/sql-createoperator.html) on a stored function, per [`CREATE OPERATOR`](https://www.postgresql.org/docs/current/sql-createoperator.html)

Issue #234, opened 2023-08-21
Adding angular distance for nn search


vgoat21:
I was wondering if it would be possible to add angular distance for nn search. I am currently relying on cosine distance as an alternative. The reason behind this is the limitation of cosine distance when dealing with small angles.

![image](https://github.com/pgvector/pgvector/assets/93329405/b433b86b-29d1-4667-bb76-a7e7302b0455)

It's apparent that cosine distance tends to produce very similar values for small angles, making it challenging to distinguish the subtle differences in vector orientations. This leads to some ambiguity in these intervals which (at least in my application) is not the most appropriate to use.

Angular distance is sensitive not only to the direction but also to the magnitude of vectors. As depicted in the diagram, it yields more distinguishable values within these intervals, thus making it easier to assess the proximity of two vectors accurately.

In my application (and most likely many others) it is crucial to include these subtle differences to get a more accurate 'closeness'.

I think this would be a great addition to a very useful extension.

phobrain:
It looks potentially feasible to add such functions in

https://github.com/pgvector/pgvector/blob/master/src/vector.c

which I was just considering for Poincare disk distances as in

https://arxiv.org/pdf/1705.08039.pdf

and since I find it empirically allows more fine-grained histogram distance orderings, I wonder naively if the Poincare sphere would be useful in your app.

Authors, is adding a new distance function as simple as l2_distance() seems in vector.c?

jkatz:
>  is adding a new distance function as simple as l2_distance() seems in vector.c

Yes, it's relatively trivial, though note that (1) optimizing how the distance calculation is handled is key to many of the performance characteristics and (2) you need to set up the appropriate [operator classes][1] to allow it to be indexable.

For this, I'd be curious if you're noticing these results currently in pgvector. For `ivfflat`, vectors are normalized before they are stored in the index, so magnitude is not a factor.

[1]: https://github.com/pgvector/pgvector/blob/master/sql/vector.sql#L257

phobrain:
Have any functions been tried and abandoned for speed reasons?

ankane:
Hi @vgoat21, thanks for the suggestion. I may include this at some point, but for now, you can do:

```sql
CREATE FUNCTION angular_distance(a vector, b vector) RETURNS float8
    LANGUAGE SQL
    IMMUTABLE
    RETURNS NULL ON NULL INPUT
    RETURN acos(1 - cosine_distance(a, b)) / pi();
``` 

[...]

PR #806, opened 2025-03-25
Added support of inner product distance for bit vectors

astanabe:
In my molecular sequence (i.e. DNA nucleotide or amino-acid sequences) analysis experience, (negative) inner product distance is often the best distance metric.
Recently, I faced to handle very large data contains over 5 billion vectors of 4096 dimensions, and, binary quantization is required.
For this analysis, I implemented inner product distance support for bit vectors.


ankane:
Hi @astanabe, thanks for the PR. However, I'd expect Hamming distance to provide better recall for this (which is already available).

A few questions:

1. How are you generating the vectors?
2. What is the recall when using exact search on the binary quantized vectors with Hamming distance and inner product (compared to the non-quantized vectors)?
3. Are there any papers that discuss this approach?

astanabe:
Consider a case where the following DNA base sequence is stored in a database:

```
DNA1 CGTATACGTATAG
```

We can create a vector by extracting 6-character-long words, shifting one character at a time, and counting their frequency.
This is similar to an n-gram indexing in full-text search.
In molecular biology, this is called a k-mer, but it is exactly the same as n-gram in full-text search.

DNA1 would look like this:

```
CGTATA 2
GTATAC 1
TATACG 1
ATACGT 1
TACGTA 1
ACGTAT 1
GTATAG 1
```

DNA consists of only four types of characters: A, C, G, and T.
Therefore, when the word length is 6 characters, there are 4096 possible words, making the vector 4096-dimensional.

Now, consider using the following as a query for searching:

```
Query CGTATAC
```

This can be broken down as follows:

```
CGTATA 1
GTATAC 1
```

In this scenario, the Hamming distance between DNA1 and the Query is 5, and the negative inner product is -2.
Furthermore, consider that the database also contains the following sequence:

```
DNA2 AGTATAC
```

For DNA2 and the Query, the Hamming distance is 1, and the negative inner product is -1.
In this search, we want to retrieve DNA1, but because DNA2 is closer when using the Hamming distance, we would need to increase the LIMIT.

In real-world molecular biology research, the database often contains a mixture of very long and short nucleotide sequences, and they are searched with short query sequences.
In this situation, when using the Hamming distance, shorter sequences that are slightly less similar can appear as closer neighbors than longer sequences that are more similar.

Jaccard distance improves upon the Hamming distance.
Sorensen distance improves upon Jaccard distance.
Simpson distance improves upon Sorensen distance.
However, the negative inner product surpasses all of these in terms of accuracy, and it is the fastest to compute in this case.
The same thing should happen when searching for short keywords or phrases in a large number of long natural language texts.

ankane:
Thanks, will spend some time getting caught up on ANN research for k-mers.

MinHash + Jaccard distance sounds like [one approach](https://academic.oup.com/nar/article/52/16/e74/7714450).

Also, would be good to see how recall compares between different distance functions / approaches on your dataset or (ideally) an open dataset.

astanabe:
GSearch and its predecessors primarily focus on searching a database that contains numerous genome (i.e. very long) sequences using genome sequences as queries, and on clustering and classifying genome sequences.
Therefore, they do not give much consideration to searching a database that contains a mixture of sequences of various lengths using shorter sequences as queries - this is the issue I am currently dealing with.

I am trying to examine recall using publicly available data, but because the database is extremely huge, it looks like it will take a considerable amount of time.

ankane:
Thanks, are there any papers that describe the approach you're taking?

Would also be good to compare with inner product and `sparsevec` (this would provide better recall than the binary approach, but likely has much worse performance).

astanabe:
This probably a new idea and there is no paper about this.

Because `sparsevec` is limited to 1000 non-zero elements, it is unable to store the data.

ankane:
> This probably a new idea and there is no paper about this.

It's probably not a good fit for pgvector if that's the case, but feel free to share when your work is public.

Unofficially, you can probably add it to an existing installation like in #234.

ankane:
Just to follow up, here's how you could add it unofficially to a database:

```sql
CREATE FUNCTION bit_negative_inner_product(bit, bit) RETURNS float8
    AS 'BEGIN RETURN -bit_count($1 & $2); END;'
    LANGUAGE plpgsql IMMUTABLE STRICT PARALLEL SAFE;

CREATE OPERATOR <#> (
    LEFTARG = bit, RIGHTARG = bit, PROCEDURE = bit_negative_inner_product,
    COMMUTATOR = '<#>'
);

CREATE OPERATOR CLASS bit_ip_ops
    FOR TYPE bit USING hnsw AS
    OPERATOR 1 <#> (bit, bit) FOR ORDER BY float_ops,
    FUNCTION 1 bit_negative_inner_product(bit, bit),
    FUNCTION 3 hnsw_bit_support(internal);
```

PR #809, opened 2025-03-26
docs: Add FAQ about pgvector feasibility for bulk distance calculations

frapercan:
This PR adds a frequently asked question and answer detailing why row-oriented databases such as Postgres (with pgvector) may not be optimal for massive distance computations.


frapercan:
Have you ever compared performance using for example sklearn compute_distances method vs this method?

Even if you mount the database in shm, with optimized postgres configuration in a HPC, there is no comparison.

I think that thousands vs thousands are already a problem. 

Anyways, thanks!!

PR #849, opened 2025-06-04
increase maximum dimension size for HNSW index

fshabashev:
Some popular embedding models, such as OpenAI's `text-embedding-3-large` have more than 2,000 dimensions. 
I propose increasing the `HNSW_MAX_DIM` from 2,000 to 4,000 to support larger embedding models.

### DIFF
--- a/src/hnsw.h
+++ b/src/hnsw.h
@@ -12,7 +12,7 @@
 #include "utils/sampling.h"
 #include "vector.h"
 
-#define HNSW_MAX_DIM 2000
+#define HNSW_MAX_DIM 4000
 #define HNSW_MAX_NNZ 1000
 
 /* Support functions */


jkatz:
This won't work as expected, as the tuple will exceed the page boundary. The technique to use for this is cast the indexed vector to `halfvec`. There's an [example here](https://jkatz05.com/post/postgres/pgvector-scalar-binary-quantization/), using the cosine distance:

```sql
CREATE TABLE documents (
    id bigint GENERATED BY DEFAULT AS IDENTITY PRIMARY KEY,
    content text,
    embedding vector(3072)
);

CREATE INDEX ON documents
    USING hnsw ((embedding::halfvec(3072)) halfvec_cosine_ops);
```

PR #863, opened 2025-06-27
Adding optional in-memory tracking of recall metrics

binarycleric:
Add recall tracking functionality to monitor the quality of approximate vector indexes without the performance overhead of exact searches. It is disabled by default and must be explicity enabled by a user. Recall tracking can also be disabled in the event of bugs or performance impacts.

Statistical recall tracking monitors index performance without expensive exact searches. The recall estimation works by counting how many tuples in the table fall within the distance of the K-th result, providing a practical approximation of search quality without the cost of brute-force comparison.

This should give users a fairly decent idea of recall without adding significant performance overhead and can be useful in observability pipelines or during development. This information is only stored in memory right now but can be persisted to disk as a future enhancement.

### Configuration

```sql
SET pgvector.track_recall = on;
SET pgvector.recall_sample_rate = 50;  -- Sample every 50th query
SET pgvector.recall_max_scan_tuples = 10000 -- How many tuples to scan before exiting.
```

### Usage

View recall summary with index names:

```sql
SELECT * FROM pg_vector_recall_summary();
```

Get raw statistics:

```sql
SELECT * FROM pg_vector_recall_stats();
```

Get statistics for a specific index:

```sql
SELECT pg_vector_recall_get(index_oid);
```

Reset stats for a specific index:

```sql
SELECT pg_vector_recall_reset(index_oid);
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,10 +1,10 @@
 EXTENSION = vector
-EXTVERSION = 0.8.0
+EXTVERSION = 0.9.0
 
 MODULE_big = vector
 DATA = $(wildcard sql/*--*--*.sql)
 DATA_built = sql/$(EXTENSION)--$(EXTVERSION).sql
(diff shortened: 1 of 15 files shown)


ankane:
Hi @binarycleric, thanks for the PR. It's an interesting idea, but since it's already possible / more flexible to track recall outside of the extension, this isn't something I'd like to include (for instance, it can be tracked asynchronously and account for different limits, search parameters, and filters).

PR #868, opened 2025-07-02
Increase sparsevec HNSW index limit to 1200 non-zero elements

### LINKED ISSUE #749: Enhancement: Increase 1,000 non-zero elements limit to 1,200 for sparsevec when creating HNSW index
HNSW Index

Current: sparsevec - up to 1,000 non-zero elements (added in 0.7.0)
Enhancement Request: sparsevec - up to 1,200 non-zero elements

ysys143:
This PR increases the maximum number of non-zero elements allowed in sparsevec for HNSW indexes from 1000 to 1200.

Users working with SPLADE reported encountering vectors with ~1100 non-zero elements. The current limit of 1000 prevents them from creating HNSW indexes on their data, forcing them to use workarounds like splitting vectors across multiple columns.

The change is straightforward - just updating `HNSW_MAX_NNZ` from 1000 to 1200 in `src/hnsw.h`. I've updated the tests to verify that vectors with 1200 elements can be indexed successfully, while 1201 elements still fail with the expected error message.

This is a small change that unblocks real-world usage without significant impact on memory or performance.

Fixes #749

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,8 @@ You can also install it with [Docker](#docker), [Homebrew](#homebrew), [PGXN](#p
 
 ### Windows
 
+> **Note:** On Windows, `maintenance_work_mem` is currently limited to 2,024MB at most (up to and including PostgreSQL 16). It might therefore take a prohibitively long time to build indices for large vector stores.
+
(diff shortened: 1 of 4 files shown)


ankane:
Hi @ysys143, thanks for the PR, but this won't work (please test changes before submitting a PR).

PR #873, opened 2025-07-09
housekeeping: improve ax and bx args by making them constants

fractalliter:
# Description

I am getting onboarded on this project and had a quick look at the bit utils file and thought to change the ax and bx reference parameters to a const.

It's a housekeeping PR anyways.


ankane:
Hi @fractalliter, thanks for the PR, but this isn't something I'd like to change (as it isn't needed).

PR #888, opened 2025-08-15
Add optimized implementations of binary_quantize

lucagiac81:
This PR adds optimized implementations of binary_quantize. 

The first commit refactors the existing code to take advantage of autovectorization by the compiler (using an inner loop for each output byte). Generation of vectorized code has been confirmed with gcc-13, gcc-14, and clang-19.

The second commit adds optimized implementations using AVX-512 and GFNI instruction sets. 
There are two components:
- Vector element comparison using AVX-512F
- Bit order inversion using GFNI (invert the bit order of the results of the previous comparison to match the bit order required by pgvector). 

The GFNI instructions (Galois Field New Instructions) were introduced in the 3rd Gen Intel® Xeon® Scalable Processors.
Note that the GFNI-based component also requires AVX-512VL for some of the instructions. This is implied in the rest of the description.

The two components are enabled based on the features detected on the system at runtime:
- If neither AVX-512F or GFNI are present, the autovectorized binary_quantize implementation is used.
- If AVX-512F is present but not GFNI, the bit order inversion is done using a lookup table approach.

The AVX-512/GFNI implementation is placed in new files (vectorutils.h/c). This follows the same pattern used for halfvec and bit vector optimized implementations.

Both approaches (autovectorization and AVX-512/GFNI) are included in this PR to better show the full picture and the benefit introduced by each of them. However, if preferred, the two commits can be submitted as individual PRs.

### Results

Setup
- AWS r7i.metal-24xl (AVX-512 and GFNI supported), Ubuntu 24.04, kernel 6.8, gcc-13
- PostgreSQL 17.4
- pgvector
  - unmodified (trunk commit this PR is based on)
  - autovectorized implementation (use first commit in this PR)
  - AVX-512/GFNI implementation (use both commits in this PR)

Workload
- VectorDBBench v1.0.5
- Cases: Performance1536D500K and Performance1536D5M (OpenAI 500k and 5M vector datasets, 1536 dimensions)
- Parameters: k = 10, quantization = bit, reranking = enabled, query concurrency = 10
- HNSW parameters: m = 16, ef_construction = 128, ef_search = 128

Observed queries/second (qps) gains over unmodified pgvector:
- With 500k vector dataset: +20.5% (autovectorized vs unmodified), +28.5% (AVX-512/GFNI vs unmodified)
- With 5M vector dataset: +10.9% (autovectorized vs unmodified), +16.8% (AVX-512/GFNI vs unmodified)
 
Recall is matched for all cases, as expected (the quantization is the same in all implementation).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ EXTVERSION = 0.8.0
 MODULE_big = vector
 DATA = $(wildcard sql/*--*--*.sql)
 DATA_built = sql/$(EXTENSION)--$(EXTVERSION).sql
-OBJS = src/bitutils.o src/bitvec.o src/halfutils.o src/halfvec.o src/hnsw.o src/hnswbuild.o src/hnswinsert.o src/hnswscan.o src/hnswutils.o src/hnswvacuum.o src/ivfbuild.o src/ivfflat.o src/ivfinsert.o src/ivfkmeans.o src/ivfscan.o src/ivfutils.o src/ivfvacuum.o src/sparsevec.o src/vector.o
+OBJS = src/bitutils.o src/bitvec.o src/halfutils.o src/halfvec.o src/hnsw.o src/hnswbuild.o src/hnswinsert.o src/hnswscan.o src/hnswutils.o src/hnswvacuum.o src/ivfbuild.o src/ivfflat.o src/ivfinsert.o src/ivfkmeans.o src/ivfscan.o src/ivfutils.o src/ivfvacuum.o src/sparsevec.o src/vectorutils.o src/vector.o
 HEADERS = src/halfvec.h src/sparsevec.h src/vector.h
 
 TESTS = $(wildcard test/sql/*.sql)
--- a/src/vector.c
+++ b/src/vector.c
@@ -21,6 +21,7 @@
 #include "utils/lsyscache.h"
 #include "utils/numeric.h"
 #include "vector.h"
+#include "vectorutils.h"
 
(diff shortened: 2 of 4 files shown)


ankane:
Hi @lucagiac81, thanks for the PR. The auto-vectorized version seems like a good trade-off, so will do some testing.

The bigger issue is `binary_quantize` should only need to be called a single time for BQ w/ re-ranking queries (for the query vector).

```sql
SELECT * FROM (
    SELECT * FROM items ORDER BY binary_quantize(embedding)::bit(3) <~> binary_quantize('[1,-2,3]') LIMIT 20
) ORDER BY embedding <=> '[1,-2,3]' LIMIT 5;
```

However, it's called once for every row in the inner query as well (20 extra times in the example above) due to the [current behavior](https://www.postgresql.org/message-id/flat/2ca5865b-4693-40e5-8f78-f3b45d5378fb%40iki.fi) of the Postgres executor (which impacts all nearest neighbor queries).

ankane:
Merged auto-vectorization in 0c9070ba82fb4e6b6997cdd76b88442175a2f8ab.  Seeing a 5-10% improvement on Linux x86-64 and 10-20% on Mac arm64 in my testing.

Thanks @lucagiac81!

lucagiac81:
Hi @ankane . Thank you very much for the fast review! If you plan to test the AVX-512/GFNI part as well, @asonje and @matt-welch can assist with any questions going forward.

PR #890, opened 2025-08-23
Use NIL for empty lists

rjuju:
Postgres standard way to check for list emptiness is to compare a pointer to NIL rather than NULL.

### DIFF
--- a/src/hnsw.c
+++ b/src/hnsw.c
@@ -130,7 +130,7 @@ hnswcostestimate(PlannerInfo *root, IndexPath *path, double loop_count,
 	Relation	index;
 
 	/* Never use index without order */
-	if (path->indexorderbys == NULL)
+	if (path->indexorderbys == NIL)
 	{
 		*indexStartupCost = get_float8_infinity();
 		*indexTotalCost = get_float8_infinity();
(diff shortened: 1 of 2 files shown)


ankane:
Thanks @rjuju!

rjuju:
thanks for merging!

ankane:
No problem. fwiw, [this comment](https://github.com/postgres/postgres/blob/REL_17_5/src/include/nodes/pathnodes.h#L1686-L1691) mentions using `NIL`, but [Postgres itself](https://github.com/postgres/postgres/blob/REL_17_5/src/backend/optimizer/plan/createplan.c#L3120) uses `NULL`.

rjuju:
Yes it's not surprising, using `NULL` is quite strong muscle memory thing in C and the additional cast done by `NIL` doesn't really add much.  It's definitely not a game changer to use NIL here, but if one day this list is changed to another structure (I think this kind of change happened quite frequently on the partitioning related structures some years ago)  you should then get a clear compiler error rather than random errors and segfaults that will make waste a bit of time at the time it breaks.  I guess that's the main reason to prefer NIL.

PR #897, opened 2025-09-22
Parallel IVF scan implementation

mariosangelis:
# Parallel IVF Scan Implementation

This Pull Request introduces a **parallel IVFFlat index scan** implementation for pgvector, extending the default `ivfflat` index scan with support for background worker parallelism.  

## Motivation

The main performance bottleneck of the default IVFFlat implementation derives from the large number of page fetches and the extensive distance computations it performs. To address this limitation, I designed a parallel IVFFlat algorithm. More specifically, I extended the default scan algorithm by parallelizing the `GetScanItems` function. The high level idea is that after computing the closest centroids, we construct an in-memory list that contains the block numbers of all the embedding pages that need to be scanned. This list is then partitioned into chunks and distributed between multiple parallel workers. Each worker independently executes the same logic as the original `GetScanItems` function, but only for its assigned chunk. This parallelization allows the index scan to benefit from concurrent processing, effectively reducing query execution time.

## Design
The parallel algorithm extends the default `ivfflat` scan in two ways:  

### 1. Index Build Phase Modification
- Extended the build process to record a per-centroid block number list. This list includes the block numbers of each embedding page assigned to the corresponding cluster centroid.
- These block lists allow us to efficiently enumerate the embedding pages that need to be scanned during query execution.  

### 2. Parallel Scan Execution
- After `GetScanLists` identifies the closest probe centroids, we revisit the corresponding centroid metapages and extract their associated block number lists. These lists are concatenated into a single in-memory list.
- The in-memory list is partitioned into chunks and distributed across multiple workers.  
- Each worker executes the same logic as the serial `GetScanItems` function, but only for its assigned pages.  
- Workers write `{distance, TID}` pairs into a per-worker shared memory region.  
- The leader process merges results incrementally by maintaining cursors into each worker’s sorted result list.  

## Benefits
- **Performance**: The parallel algorithm leverages available parallel workers (`max_parallel_workers_per_gather`, `max_worker_processes`) to speed up queries.  We can achieve up to **~4x** speedup using 4 parallel workers.
- **Compatibility**: The algorithm falls back to the serial execution scan when no workers are available.  

## Testing
✅ All tests pass:  
```bash
make installcheck        # regression tests
make prove_installcheck  # TAP tests
```

## Performance Evaluation

For the evaluation of the optimized IVFFlat index algorithm, I used the [ANN-Benchmarks framework](https://github.com/erikbern/ann-benchmarks), which enables the execution of top-K workloads on publicly available datasets.  

I compared three implementations:
- **Default pgvector IVFFlat**
- **Default pgvector HNSW**
- **Optimized parallel IVFFlat**

### Results
- The **parallel IVFFlat** consistently outperforms the default IVFFlat implementation across all tested datasets.
- For the **GloVe dataset**, the parallel implementation achieves a **3.5× speedup in query throughput**, for recall levels > 0.9.  
- The index build phase introduces negligible overhead compared to the default implementation.  

Screenshot from 2025-09-22 16-03-04

## Summary
This Pull Request introduces a parallel IVFFlat index scan algorithm, aiming to improve the index scan query performance on large `ivfflat` indexes. It provides significant speedups (**up to ~4× for 4 background workers**) by parallelizing distance computations across multiple workers.  

Feedback and review are very welcome! 🚀

### DIFF
--- a/src/ivfbuild.c
+++ b/src/ivfbuild.c
@@ -19,6 +19,7 @@
 #include "tcop/tcopprot.h"
 #include "utils/memutils.h"
 #include "vector.h"
+#include <math.h>
 
 #if PG_VERSION_NUM >= 140000
 #include "utils/backend_progress.h"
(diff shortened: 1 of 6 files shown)


ankane:
Hi @mariosangelis, thanks for the PR. It'd be nice to support parallel index scans for IVFFlat, but it should use the [index AM](https://www.postgresql.org/docs/current/index-functions.html) to do so. Last time I tried this, the Postgres planner would get the cost estimate, but never chose to use the index ([parallel-index-scan3 branch](https://github.com/pgvector/pgvector/compare/parallel-index-scan3)), so this would need to be addressed in Postgres first.

Also, for PRs, please keep changes to a minimum (no style changes).

PR #921, opened 2025-11-07
Add .DS_Store to .gitignore to prevent tracking of macOS system files

matheusallvarenga:


ankane:
Hi @matheusallvarenga, users can use a global gitignore for this.

PR #927, opened 2025-11-19
Support sq16 in hnsw access method with vector_cosine_ops

zthhhhh:
Hi pgvector teams,

In this pull request, I would like to add a new operator class family for hnsw access method, and it is used to enhance the performance when conducting vector search based on cosine distance.
# Theory
In the calculation of vector cosine distance, what we focus on is the direction of each vector. That is to say, If a coefficient is multiplied in each dimension of the vector, this will not lead to the final calculation result of the cosine distance, without considering the loss of accuracy. So we can use an int16 to represent the value of each dimension and use the avx512 instruction to optimize the distance calculation function.

This has already been supported and verified to be feasible in some open-source databases, such as mariadb mhnsw index. 
relevant code:
https://github.com/MariaDB/server/blob/main/sql/vector_mhnsw.cc#L898

# Performance
I run ann-benchmark in my test machine with 'dbpedia-openai-1000k-angular' dataset, setup enough large `shared_buffers` and `maintenance_work_mem` to eliminate the impact of disk I/O during index building and vector search phase, and it shows promising results. Under the same recall rate, we achieved a higher qps. 
image

Moreover, the construction time has been increased from 1517.7 seconds to 936.4 seconds, using the same index build parameter.

In addition, I noticed that hnsw will have a significant performance regression when memory is limited, there are some solutions  for this question, such as PQ. Would the pgvector team like have plans to support them?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,6 +32,12 @@ endif
 # - Clang (could use pragma instead) - https://llvm.org/docs/Vectorizers.html
 PG_CFLAGS += $(OPTFLAGS) -ftree-vectorize -fassociative-math -fno-signed-zeros -fno-trapping-math
 
+ARCH := $(shell uname -m)
+ifeq ($(ARCH),x86_64)
+PG_CFLAGS += -mavx512f -mavx512dq  -mavx512vnni
+COMPILE.c.bc = $(CLANG) -Wno-ignored-attributes $(BITCODE_CFLAGS) $(CPPFLAGS) -flto=thin -emit-llvm -c -mavx512f -mavx512dq  -mavx512vnni -ftree-vectorize -fassociative-math -fno-signed-zeros -fno-trapping-math
+endif
+
 # Debug GCC auto-vectorization
 # PG_CFLAGS += -fopt-info-vec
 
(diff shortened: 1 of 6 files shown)


ankane:
Hi @zthhhhh, thanks for the PR.

It'd be good to see how performance compares to the existing quantization options (half-precision indexing and binary quantization with re-ranking), as well as int8 quantization.

Also, please don't reference code from other vector libraries if planning to contribute to pgvector (for licensing).

Re PQ: I think per-vector quantization is likely better for most users since it doesn't require a training step. For limited memory, binary quantization with re-ranking can often be used.

zthhhhh:
Thank you for your response! I’ll run more detailed benchmarks to verify the performance characteristics. If the implementation in my PR doesn’t consistently outperform halfvec, it may indeed not be a worthwhile approach. Because their compression ratios are the same.

Regarding vector quantization, I’ve extended the ivfflat access method to support different quantization strategies within the handler while maintaining full backward compatibility with existing ivfflat indexes. As a proof of concept, I implemented IVF-RABITQ (with cosine distance) using an operator class on top of this extension, and it’s working successfully.

If this is of interest, I wanna to share more details, and would appreciate any guidance on how to contribute these implements to the open-source pgvector community.

Thanks again!

ankane:
Closing this out, but happy to revisit when there's data.

Re RaBitQ / other quantization strategies: Feel free to create a PR with the proof of concept and the same benchmarking data requested above.

PR #937, opened 2025-12-11
Improve vector calculations performance

bartoszjuja:
The dependency chain was a bottleneck in some of the vector calculations - each iteration must wait for the previous accumulation to complete. Modern CPUs can execute multiple independent basic operations in parallel even on a single core. This means that partially unrolling the loop allows the CPU to run multiple addition/multiplication commands at the same time.

Testing shows that using 4 independent accumulators results in a 25-30% performance increase for vectors with a length of 1000.

### DIFF
--- a/src/vector.c
+++ b/src/vector.c
@@ -549,13 +549,30 @@ halfvec_to_vector(PG_FUNCTION_ARGS)
 VECTOR_TARGET_CLONES static float
 VectorL2SquaredDistance(int dim, float *ax, float *bx)
 {
-	float		distance = 0.0;
+	/* Use 4 independent accumulators to break dependency chains */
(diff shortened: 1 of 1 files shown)


ankane:
Hi @bartoszjuja, this isn't needed with auto-vectorization / doesn't help performance.

bartoszjuja:
> Hi @bartoszjuja, this isn't needed with auto-vectorization / doesn't help performance.

@ankane seems very hasty of you to dismiss my PR without actually reading into it or asking for more data.
It does help performance because auto vectorization only makes use of SIMD, it doesn't unblock instruction level parallelism. I also mentioned I've run tests and saw up to 30% improvement.

PR #942, opened 2025-12-31
Rename array_isspace into scanner_isspace in the comment

ivan-v-kush:
`array_isspace` was removed from PG in https://github.com/postgres/postgres/commit/ae6d06f09684d8f8a7084514c9b35a274babca61

Currently `scanner_isspace` is used everywhere

### DIFF
--- a/src/vector.c
+++ b/src/vector.c
@@ -130,7 +130,7 @@ InitVector(int dim)
 }
 
 /*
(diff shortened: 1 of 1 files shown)


ankane:
Hi @ivan-v-kush, thanks for sharing. Updated each type to use `scanner_isspace` for Postgres 17+ in the commit above.

PR #943, opened 2026-01-04
Fix a detail in the tap test case.

zthhhhh:
Hi pgvector team,

Recently, while writing regression tests for my patch, I discovered an issue. In the test script 005_ivfflat_query_recall.pl, ivfflat vector indexes are created separately for l2_distance, ip_distance, and cosine_distance. However, only the recall results for l2_distance are verified against expected values. This appears to be an oversight, although it likely has minimal impact, I still believe it's worth fixing.


ankane:
Hi @zthhhhh, thanks for reporting. Simplified the test case in the commit above.

Issue #380, opened 2023-12-16
Add json and jsonb inout casts


Komzpa:
Hello,

I started using pgvector and found that I can't cast json / jsonb expressions into vectors although syntax is the same.
The fix is to create them using inout. I expect both sides to work, so this piece of code needs to land in proper spot in the SQL files:

```
create cast (json as vector) with inout;
create cast (vector as json) with inout;
create cast (jsonb as vector) with inout;
create cast (vector as jsonb) with inout;
```

In action:

```
16:13:13 [gis] > select '[1,2,3]'::json::vector;
ERROR:  42846: cannot cast type json to vector
LINE 1: select '[1,2,3]'::json::vector;
                              ^
LOCATION:  transformTypeCast, parse_expr.c:2657
Time: 0.612 ms
16:15:46 [gis] > create cast (json as vector) with inout;
CREATE CAST
Time: 0.508 ms
16:15:59 [gis] > select '[1,2,3]'::json::vector;
┌─────────┐
│ vector  │
├─────────┤
│ [1,2,3] │
└─────────┘
(1 row)

Time: 0.406 ms

```

Thanks for getting this extension for us!

Komzpa:
Hey, this got closed really fast - will you be willing to accept a pull request about this? It's quite literally four lines of code that I posted above so I am not sure about your concern of being not common enough. I met this when calling OpenAI embeddings generation directly from SQL using [pgsql-http](https://github.com/pramsey/pgsql-http).

jkatz:
@Komzpa Do you have any benchmarks to compare your implementation over what Andrew suggested (`jsonb::text::vector`)? It'd be good to see if it reduces overhead.

Komzpa:
@jkatz it connects the in function of the new type to out function of old type. The main benefit of such arrangement is that one day you may decide that you want to write a specialized high-performance converter from json(b) into vector and all the codebase will immediately benefit for that, as it will not get stuck with ::text:: inbetween. Non-scientific oneliner benchmark shows 12% faster conversion.

```

20:18:08 [kom] > select count(embedding::text::jsonb) from cache_openai_embeddings;
...
(1 row)

Time: 1946,248 ms (00:01,946)


20:18:31 [kom] > select count(embedding::jsonb) from cache_openai_embeddings;
....
(1 row)

Time: 1702,943 ms (00:01,703)

```

PR #944, opened 2026-01-05
Add jsonb_to_vector and jsonb_to_halfvec cast functions

jkatz:
This commit adds the ability to cast JSONB arrays directly into vector and halfvec vector types, for example:

```sql
SELECT '[1,2,3]'::jsonb::vector(3);
SELECT '[1,2,3]'::jsonb::halfvec(3);
```

There are several TODOs if the overall idea of the patch is acceptable:

* Determine if the method should be applied to `sparsevec` (and possibly bit vectors, but maybe a separate discussion given `bit` is a core type).
* Update README and other docs with examples.
* For discussion: to go from `vector` to `jsonb`, one must cast to an array (`to_jsonb(vector::real[])`) otherwise the value is stored as a string. It may be helpful to have a `vector_to_jsonb` function similar to how PostgreSQL has `array_to_json`.

------
## Motivation

There are several vector data sources that come from JSON documents, whether the embeddings are contained in them or they're being transformed from other sources, such as from data lake files. Additionally, some cases prefer not to duplicate the vector data between the JSON file and a separate vector column, though are fine to use the vector as an expression in a HNSW/IVFFlat index. The [previous discussion](https://github.com/pgvector/pgvector/issues/380#issuecomment-1858844472) concluded that the cast from `jsonb::text::vector` would work, but for cases with bulk imports or transformations, this adds nontrivial overhead.

## Testing

The tests show how the `jsonb_to_vector` functions performs compared to `jsonb::text::vector` cast. This was executed as `k=10` exact-nearest neighbor queries on a dataset of 100,000 vectors that all fit into memory. Each test was run until 50 or 500 transactions was completed, and the average of these transactions taken. The times are in milliseconds.

I went through a few different variations of the tests including:

1. Baselining against a regular, non-casted K-NN query
1. Casting from JSONB to an array to a vector type. This was 10x slower overall than the other results, thus not printing.
1. Casting from JSONB to text to a vector type
1. Casting directly from JSONB to a vector type

The below show the results from the last two tests

**`jsonb::text::vector`**

```sql
SELECT id,
    (embedding_jsonb::text::vector(1536)) <=>
        (SELECT embedding_vector FROM data WHERE id = ?) as distance
FROM data
ORDER BY distance
LIMIT 10;
```

**`jsonb_to_vector`** (similar for `jsonb_to_halfvec`)

```sql
SELECT id,
    (embedding_jsonb::vector(1536)) <=>
        (SELECT embedding_vector FROM data WHERE id = ?) as distance
FROM data
ORDER BY distance
LIMIT 10;
```

### Results

Most tests showed a direct jsonb to vector cast had close to a 20% speedup over the `jsonb::text::vector` method.

#### vector

Dimension | `jsonb::text::vector` | `jsonb_to_vector` | % Speedup
-- | -- | -- | --
128 | 282.594 | 237.999 | 15.8%
768 | 2343.711 | 1881.062 | 19.7%
1536 (external) | 4566.775 | 3723.697 | 18.5%
1536 (plain) | 3139.608 | 2532.739 | 19.3%

#### halfvec

Dimension | jsonb::text::halfvec | jsonb_to_halfvec | % Speedup
-- | -- | -- | --
128 | 290.947 | 237.163 | 18.5%
768 | 1623.813 | 1295.505 | 20.2%
1536 (external) | 4591.033 | 3695.503 | 19.5%
1536 (plain) | 3139.492 | 2509.346 | 20.1%

### DIFF
--- a/sql/vector.sql
+++ b/sql/vector.sql
@@ -146,6 +146,9 @@ CREATE FUNCTION array_to_vector(double precision[], integer, boolean) RETURNS ve
 CREATE FUNCTION array_to_vector(numeric[], integer, boolean) RETURNS vector
 	AS 'MODULE_PATHNAME' LANGUAGE C IMMUTABLE STRICT PARALLEL SAFE;
 
+CREATE FUNCTION jsonb_to_vector(jsonb, integer, boolean) RETURNS vector
+	AS 'MODULE_PATHNAME' LANGUAGE C IMMUTABLE STRICT PARALLEL SAFE;
+
 CREATE FUNCTION vector_to_float4(vector, integer, boolean) RETURNS real[]
 	AS 'MODULE_PATHNAME' LANGUAGE C IMMUTABLE STRICT PARALLEL SAFE;
 
@@ -168,6 +171,9 @@ CREATE CAST (double precision[] AS vector)
 
 CREATE CAST (numeric[] AS vector)
 	WITH FUNCTION array_to_vector(numeric[], integer, boolean) AS ASSIGNMENT;
+
+CREATE CAST (jsonb AS vector)
+	WITH FUNCTION jsonb_to_vector(jsonb, integer, boolean) AS ASSIGNMENT;
 
 -- vector operators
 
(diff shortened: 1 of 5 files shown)


ankane:
Thanks @jkatz. I'm hesitant about including a direct cast for this, since I don't think it's super common and the existing cast is fairly performant.

jayeshds:
wqess

sagitrs (changes requested):
-

helmize1978 (approved):

PR #948, opened 2026-01-13
Add RISC-V64 architecture support in Makefile

MengMengDaXiaoJi:
On RISC-V64 platform, '-march=native' is not supported, so we should use '-march=rv64gc' instead.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,6 +27,11 @@ ifneq ($(filter ppc64%, $(shell uname -m)), )
 	OPTFLAGS =
 endif
 
+# RISC-V64 doesn't support -march=native
+ifeq ($(shell uname -m), riscv64)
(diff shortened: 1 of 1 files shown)


ankane:
Hi @MengMengDaXiaoJi, thanks for the PR. Let's skip passing `-march` for RISC-V64 (similar to PowerPC).

Edit: It should also be possible to simplify the check to `ifeq ($(shell uname -m), riscv64)`

MengMengDaXiaoJi:
Thanks, I will check and modify my shell script.

MengMengDaXiaoJi:
Hi @ankane ,  thanks for your kind suggestion, I modified it.

ankane:
Great, thanks @MengMengDaXiaoJi!

PR #964, opened 2026-03-06
Optimize sparsevec L1/L2 distance traversal

cxmcc:
## Summary
- Optimize `SparsevecL2SquaredDistance` and `sparsevec_l1_distance` in `src/sparsevec.c` by switching from the previous nested-scan pattern to a linear two-pointer merge traversal over sorted indices.
- Keep `SparsevecInnerProduct` logic unchanged.

## Why this can improve performance
Sparse vectors store sorted indices. A two-pointer merge traversal can reduce control-flow overhead by walking both vectors in a single pass and handling leftovers in straight-line loops.

For L1/L2 distance kernels, this improves the mismatch-heavy path that dominates sparse workloads while preserving exact results.

## Latest benchmark results (PostgreSQL, L1/L2 only)
Bench setup:
- PostgreSQL 14, fresh temp cluster per variant
- `dim = 1,000,000`
- synthetic deterministic sparse data
- 5 measured runs per query (`EXPLAIN (ANALYZE, TIMING OFF)`)
- reverse-order run (`current` first, then `old`) to reduce order/cache bias

Speedup reported as `old / current`:

| nnz | L2 squared speedup | L1 speedup |
| ---: | ---: | ---: |
| 50  | 1.257x | 1.272x |
| 100 | 1.253x | 1.276x |
| 200 | 1.085x | 1.097x |
| 400 | 1.079x | 1.061x |

## Validation
- `make -j4`
- `make installcheck` (all regression tests passed)

### DIFF
--- a/src/sparsevec.c
+++ b/src/sparsevec.c
@@ -816,40 +816,38 @@ SparsevecL2SquaredDistance(SparseVector * a, SparseVector * b)
 	float	   *ax = SPARSEVEC_VALUES(a);
 	float	   *bx = SPARSEVEC_VALUES(b);
 	float		distance = 0.0;
-	int			bpos = 0;
+	int			i = 0;
+	int			j = 0;
 
-	for (int i = 0; i < a->nnz; i++)
+	while (i < a->nnz && j < b->nnz)
 	{
 		int			ai = a->indices[i];
-		int			bi = -1;
+		int			bi = b->indices[j];
 
-		for (int j = bpos; j < b->nnz; j++)
+		if (ai == bi)
 		{
-			bi = b->indices[j];
+			float		diff = ax[i] - bx[j];
 
-			if (ai == bi)
-			{
-				float		diff = ax[i] - bx[j];
-
-				distance += diff * diff;
-			}
-			else if (ai > bi)
-				distance += bx[j] * bx[j];
-
-			/* Update start for next iteration */
-			if (ai >= bi)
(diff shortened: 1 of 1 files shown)


kitturmanasi-lgtm (approved):

ankane:
Hi @cxmcc, thanks for the PR! Will spend some time benchmarking this. A few questions:

1. Can you share more about the benchmark setup for the numbers above (number of vectors, query, any indexes)?
2. Did you try the approach with inner product? If so, what were the results?

ankane:
I'm seeing a regression on Linux and a speedup on Mac with seq scans, 100k vectors, 32k dimensions, and nnz = uniform(100, 1000):

Platform | L2 | L1 | IP ([branch](https://github.com/pgvector/pgvector/compare/sparsevec-ip))
--- | --- | --- | ---
Linux x86-64 | 1.00 | 0.98 | 0.97
Linux aarch64 | 0.95 | 0.97 | 0.97
Mac arm64 | 1.15 | 1.31 | 1.31

I also tried a static nnz, but the results were similar.

cxmcc:
Thanks for review and the benchmarks and for catching the platform differences.

My setup used more sparse data and with/without indexes. Under those conditions, I consistently saw improvements for L1/L2. I did try inner product, but didn’t see improvement, so I didn't include IP in the original patch.

Given the lack of gains (and regression) on x86/amd64, it makes sense to close this PR.

Issue #963, opened 2026-03-06
Question: Binary initialisation


Grunthos:
I have a need to create sparse vector from binary source. Specifically:

- ( indexes int[] , values int[] )   -- Using an int for values at each index
- ( indexes int[], values float[] )  -- Using a float for values at each index
- ( indexes int[], value )  -- Using a fixed value at each index (int/float)

I apologise if this has already been discussed, but would this be an acceptable/useful additional feature in your view? I am happy to do the leg work to make this happen.

ankane:
Hi @Grunthos, the second case is already possible if you use the binary format ([relevant code](https://github.com/pgvector/pgvector/blob/v0.8.2/src/sparsevec.c#L504)). I don't think additional types of values are common enough to support right now.

Grunthos:
Hi @ankane,

I think you may have misunderstood my (sparsely stated) intent (or I have misunderstood the answer). 

I meant those to be SQL functions; ie. something which I can use in queries. The link you posted was (I think) fow a wire protocol decoder?

Something that would support:
```
CREATE FUNCTION sparsevec_from_arrays(indices int[], values float4[])
RETURNS sparsevec
```

ankane:
Ah, gotcha. You can currently create a custom function for it:

```sql
CREATE FUNCTION sparsevec_from_arrays(indices int[], value float4[], dimensions int) RETURNS sparsevec AS $$
    SELECT ('{' || (SELECT string_agg(indices[i] || ':' || value[i], ',') FROM generate_series(1, array_length(indices, 1)) i) || '}/' || dimensions)::sparsevec;
$$ LANGUAGE SQL IMMUTABLE STRICT PARALLEL SAFE;
```

```sql
SELECT sparsevec_from_arrays('{1,3,5}', '{2,4,6}', 6);
```

Can you share more about the use case, and why it's important to have the functionality in the database vs. the client libraries?

Grunthos:
Hi @ankane ,

Thanks for the prompt replies!

The problem with the above approach is the creation and destruction of massive strings (in my case vectors with in the worst case up up to 8k values). Mostly they are small.

The use case is as follows:

- We build arrays of what are basically attribute lists for objects.
  - This is our int[] parameter.
  - Those attribute lists are of varying length but consist of a list of integer IDs.
- We also have "weights" for those attributes stored/derived elsewhere; many objects share attributes so we weight them in one place and can easily generate a per-object list of weights based on the attributes.
[...]

PR #966, opened 2026-03-08
Add support for sparsevec(int[], {real|double|int|numeric}[], int) to sparsevec

Grunthos:
Creates a sparevec from binary data in arrays

Also added testbed cases and attempted to mimic exisiting coding style.

Performance comparison approximately 30x faster (or more) than array->text->sparsevec:

```
sparsevec constructor benchmark  (N=100000 iterations)

A: input already available as text or array (conversion cost to sparsevec only)
C: starting from arrays via text intermediate (C1) vs direct constructor (C2)

Scenario             n=10     n=100    n=500    n=1000 (n=number of entries in sparsevec)  (ns/call)
-------------------- -------- -------- -------- --------
A1 text->sv               600     5423    27391    55141
A2 real[]->sv             197      707     2671     5678
A3 integer[]->sv          220      739     2869     6019
A4 float8[]->sv           300      733     3080     6018
A5 numeric[]->sv          752     6067    29451    59271

C1 arr->txt->sv          6081    26904   115771   227328
C2 arr->sv (direct)       220      706     2770     5563
-------------------- -------- -------- -------- --------
```

### DIFF
--- a/sql/vector.sql
+++ b/sql/vector.sql
@@ -736,6 +736,18 @@ CREATE FUNCTION l2_norm(sparsevec) RETURNS float8
 CREATE FUNCTION l2_normalize(sparsevec) RETURNS sparsevec
 	AS 'MODULE_PATHNAME', 'sparsevec_l2_normalize' LANGUAGE C IMMUTABLE STRICT PARALLEL SAFE;
 
+CREATE FUNCTION sparsevec(int[], real[], int) RETURNS sparsevec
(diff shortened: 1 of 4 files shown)


Grunthos:
P.S. Happy to add the arithmetic ops (+, -, *) between scalars and sparsvec and two sparsevecs if desired

ankane:
Hi @Grunthos, see https://github.com/pgvector/pgvector/issues/963#issuecomment-4009666867 (still have the same thoughts as a few days ago).

Grunthos:
No problem, it's here if people want more complete sparsevec support and a 97% performance improvement for construction over the recommended method.

PR #978, opened 2026-04-14
Fix reservoir sampling and optimize normalization in ivfflat build

lychen0618:
This PR includes two changes to the `ivfflat` build path.

### 1. Fix reservoir sampling

`reservoir_get_next_S()` should use the total number of rows seen so far, rather than the current number of retained samples.

This matches the usage in PostgreSQL, for example in `src/backend/commands/analyze.c:acquire_sample_rows`, and is also consistent with the [reservoir sampling paper](https://dl.acm.org/doi/epdf/10.1145/3147.3165): "Let us denote the number of records processed so far by t."

### 2. Optimize sample normalization

For spherical distance, normalization is unnecessary for rows that are not retained by reservoir sampling.

This change keeps the normalization check during sampling, but defers the actual normalization until after sampling completes, so only retained rows are normalized.

### Dataset

100,000 rows from SIFT

### Script

```sql

DROP TABLE IF EXISTS items;

CREATE TABLE items (id int, embedding vector(128));

COPY items(id, embedding) FROM '/tmp/pgvector/sift_base_10w.csv' CSV;

CREATE INDEX ON items USING ivfflat (embedding vector_ip_ops) WITH (lists = 100);
```

### Results

**Without the second change**

**With the second change**

### DIFF
--- a/src/ivfbuild.c
+++ b/src/ivfbuild.c
@@ -62,15 +62,13 @@ AddSample(Datum *values, IvfflatBuildState * buildstate)
 	Datum		value = PointerGetDatum(PG_DETOAST_DATUM(values[0]));
 
 	/*
-	 * Normalize with KMEANS_NORM_PROC since spherical distance function
-	 * expects unit vectors
+	 * Check with KMEANS_NORM_PROC that the value can be normalized
+	 * since spherical distance function expects unit vectors
 	 */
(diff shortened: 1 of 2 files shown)


ankane:
Thanks @lychen0618, good finds! All three commits have been merged.

PR #988, opened 2026-06-15
Fix HNSW index corruption during high INSERT and VACUUM concurrency

bhagyesh-c:
## Summary

Under high concurrent DML load, HNSW indexes can become corrupted during `VACUUM`. This leads to persistent failures (`different vector dimensions` errors) in subsequent vacuums and and queries which scan the corrupted element. 

The corruption happens when a concurrent `INSERT` promotes a new element to the global entry point during vacuum's first phase (`RemoveHeapTids`). Because `VACUUM` caches a stale view of the entry point, it incorrectly assigns this newly promoted entry point as the `highestPoint` (which should strictly be a non-entry point). This ultimately causes vacuum to skip repairing the entry point, leaving behind a dangling edge.

## Reproduction

This race condition can be reliably reproduced using a GDB script that pauses the vacuum worker during the critical window while running concurrent inserts to force a promotion. 

* **Reproduction Script:** [reproduction.sh](https://gist.github.com/bhagyesh-c/81141580b22261d7eacb20050d4fee77)

It can also be reproduced by heavy concurrent insert/update workloads running alongside vacuum.

## Root Cause

The corruption occurs in three steps:

1. **Phase 1 (`RemoveHeapTids`):** Vacuum scans pages to find a `highestPoint` candidate. It evaluates elements using a cached, stale `entryPoint`. If a concurrent insert promotes a new element to the entry point, `blkno == entryPoint->blkno` evaluates to `false`. Vacuum mistakenly selects this new entry point as the `highestPoint`.
2. **Phase 2 (`RepairGraphEntryPoint`):** Vacuum fetches the true latest entry point (the new element) and calls:
   `RepairGraphElement(vacuumstate, entryPoint, highestPoint);`
   Because `entryPoint` and `highestPoint` are now the exact same element, the safety check inside `RepairGraphElement` triggers and exits immediately. The entry point is left unrepaired and still contains edges pointing to deleted elements.
3. **Phase 3 (`MarkDeleted`):** When these elements are physically zeroed out, the index is left with a dangling edge pointing to a 0-dimension vector.

## Fix

1. **Add a Fallback:** Introduce a `fallbackPoint` inside `RemoveHeapTids` to track a secondary candidate for the highest point.
2. **Shift Synchronization:** Move the `HNSW_UPDATE_LOCK` unlock sequence earlier, from the start of `RepairGraph` to the end of `RemoveHeapTids`. This ensures all in-flight concurrent inserts commit before we verify the entry point.
3. **Safe Demotion:** . We then check the latest `entryPoint`. If it matches our scanned `highestPoint`, we demote `highestPoint` to the `fallbackPoint` so that Phase 2 operates on a valid non-entry point.

## Verification
* The provided reproduction script completes successfully and reports no index corruption.

### DIFF
--- a/src/hnswvacuum.c
+++ b/src/hnswvacuum.c
@@ -37,6 +37,8 @@ RemoveHeapTids(HnswVacuumState * vacuumstate)
 {
 	BlockNumber blkno = HNSW_HEAD_BLKNO;
 	HnswElement highestPoint = &vacuumstate->highestPoint;
+	HnswElementData fallbackPointData;
+	HnswElement fallbackPoint = &fallbackPointData;
 	Relation	index = vacuumstate->index;
 	BufferAccessStrategy bas = vacuumstate->bas;
 	HnswElement entryPoint = HnswGetEntryPoint(vacuumstate->index);
@@ -44,10 +46,13 @@ RemoveHeapTids(HnswVacuumState * vacuumstate)
 
 	/* Store separately since highestPoint.level is uint8 */
 	int			highestLevel = -1;
+	int			fallbackLevel = -1;
 
-	/* Initialize highest point */
+	/* Initialize highest point and fallback point */
 	highestPoint->blkno = InvalidBlockNumber;
 	highestPoint->offno = InvalidOffsetNumber;
+	fallbackPoint->blkno = InvalidBlockNumber;
+	fallbackPoint->offno = InvalidOffsetNumber;
 
 	while (BlockNumberIsValid(blkno))
 	{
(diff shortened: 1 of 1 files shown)


ankane:
Hi @bhagyesh-c, thanks for the report and PR. I've confirmed the issue, and the overall approach looks good. I've created [a branch](https://github.com/pgvector/pgvector/compare/hnsw-vacuum-fix) that shuffles the logic a bit. Let me know your thoughts.

bhagyesh-c:
Hi @ankane, thanks for the quick response. I've reviewed the changes in your branch, the fix looks good and cleaner than the original. I've also verified it against the reproduction test.

ankane:
Thanks for the quick review. Applied the changes in the commit above. Also evaluating ways to make indexes more robust to this. Once that's complete and more testing has been done, will release 0.8.3.

ankane:
Added two more commits around this:

1. 98dc4aa6d0ebc39c2160b35b509fb68f7d67e28d - a vacuum pass to confirm the graph was repaired before marking tuples as deleted (very quick to perform)
2. 7b72aeeff3b5355a00214fc6112dfb7e4dc0f796 - a check when trying to an load element

ankane:
Just released 0.8.3 with the fix. For completeness, this has been present since the initial HNSW release.

PR #997, opened 2026-07-07
Drop -march=native on aarch64 GCC to avoid pac-ret miscompile

dbroudy:
GCC on aarch64 miscompiles the pac-ret return sequence when -march=native enables the pauth feature and the server is built with -mbranch-protection (as PostgreSQL 19 is on Debian). The function epilogue authenticates the return address twice ("autiasp" then "retaa"); the second authentication fails and poisons the program counter, so the backend segfaults on module load (LOAD 'vector' / CREATE EXTENSION vector), before any SQL object is created.

Drop -march=native when the compiler is GCC, the server uses -mbranch-protection=standard|pac-ret, and native enables pauth. This keeps pointer-authentication hardening intact and only forgoes native micro-tuning on affected GCC builds; clang and other compilers are unaffected.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,6 +32,25 @@ ifeq ($(shell uname -m), riscv64)
 	OPTFLAGS =
 endif
 
+PG_CONFIG ?= pg_config
+
+# GCC on aarch64 miscompiles the pac-ret return sequence when -march=native
+# enables the pauth feature and the server is built with -mbranch-protection:
+# the epilogue authenticates the return address twice ("autiasp" then "retaa"),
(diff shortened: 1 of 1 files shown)


ankane:
Hi @dbroudy, thanks for the PR. I think it'd be better to get this fixed in GCC (or use a patched version if it already is) rather than try to detect it in pgvector, but happy to revisit this if it becomes a common issue in the future.

Also, I wasn't able to reproduce it (as `-march=native` didn't enable `pauth` on the three machines I tried).

PR #1012, opened 2026-08-07
Fix parallel aggregate failure on empty sets in vector_combine.

jainvaibhavjain:
PostgreSQL parallel aggregation uses `INITCOND='{0}'` as the initial working state, resulting in a 0-dimensional float array when a table (or sequential scan shard) contains no rows. When multiple background workers yield empty shards, `vector_combine` erroneously evaluated `CheckDim(0)` on this initialization state, throwing error `vector must have at least 1 dimension` error and failing the query.

This modifies `vector_combine_impl` to conditionally bypass `CheckDim(dim)` only when both incoming parallel states are completely empty (`n == 0.0`), while fully preserving dimensionality checks against malformed inputs in all other code paths.

Also includes a dedicated TAP test to guarantee safety under forced parallel execution limits.

### DIFF
--- a/src/vector.c
+++ b/src/vector.c
@@ -1233,7 +1233,9 @@ vector_combine(PG_FUNCTION_ARGS)
 	{
 		n = n2;
 		dim = STATE_DIMS(statearray2);
-		CheckDim(dim);
+		if (n > 0.0) {
+			CheckDim(dim);
+		}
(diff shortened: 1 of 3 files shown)


ankane:
Hi @jainvaibhavjain, thanks for the PR! Pushed a version of this in the commit above. It looks like this was broken in the previous release as part of hardening efforts.

PR #1016, opened 2026-08-12
fix sparsevec_to_vector: add index bounds check to prevent out‑of‑bou…

lyinglyingl:
…nds access


ankane:
Hi @lyinglyingl, I don't see how `indices` would contain an out-of-bounds index, but if you have a reproduction script, please send an email to the [address here](https://github.com/pgvector/pgvector?tab=security-ov-file).

Edit: This could still make sense for hardening.

ankane:
Merged a version of this in the commit above as part of hardening.

PR #1023, opened 2026-09-07
Fix zero random value when choosing HNSW level

cuishuang:
`HnswInitElement()` currently chooses the HNSW level with:

```c
int level = (int) (-log(RandomDouble()) * ml);
```

`RandomDouble()` can return zero on supported PostgreSQL versions:

- PostgreSQL 15+: `pg_prng_double(&pg_global_prng_state)` returns values in `[0, 1)`.
- PostgreSQL 13/14: `random() / MAX_RANDOM_VALUE` can produce zero.

When this happens, the calculation becomes:

```text
log(0)       -> -Infinity
-log(0)      -> +Infinity
(int) +Inf   -> implementation-defined / non-portable result
```

The resulting level may then be converted to `uint8`, so this is a correctness and portability issue.

Retry the random-number generation when it returns exactly zero before calling `log()`:

```c
do
    uniform = RandomDouble();
while (uniform == 0.0);
```

This preserves the existing distribution and PostgreSQL 13/14 endpoint behavior. In particular, it avoids using `1.0 - RandomDouble()`, which would change the distribution and endpoint semantics.


ankane:
Hi @cuishuang, thanks for reporting and the PR! Merged a version of this in the commit above that avoids looping.

PR #1030, opened 2026-09-19
Added vector_dims for sparsevec

arif-js:
`vector_dims` works for `vector` and `halfvec`, but there is no way to get the dimensions of a `sparsevec`:

    SELECT vector_dims('[1,2,3]'::vector);   -- 3
    SELECT vector_dims('[1,2,3]'::halfvec);  -- 3
    SELECT vector_dims('{1:1,2:2,3:3}/3'::sparsevec);
    -- ERROR:  function vector_dims(sparsevec) does not exist

Casting to `vector` is not a workaround. `sparsevec` supports up to 1,000,000,000 dimensions, while `vector` is limited to 16,000, so for a large sparse vector the cast fails:

    SELECT '{}/20000'::sparsevec::vector;
    -- ERROR:  vector cannot have more than 16000 dimensions

With this change:

    SELECT vector_dims('{}/1000000000'::sparsevec);  -- 1000000000
    SELECT vector_dims('{1:3,4:7}/100'::sparsevec);  -- 100, not the 2 non-zeros

Implemented as `sparsevec_vector_dims`, following the existing `halfvec_vector_dims`.

Added:
- `vector_dims(sparsevec)` to `sql/vector.sql` and the `0.8.6--0.8.7` upgrade script
- tests in `test/sql/sparsevec.sql` and `test/expected/sparsevec.out`
- README and CHANGELOG entries

`make installcheck` passes (14/14).


ankane:
Hi @arif-js, can you share the use case for this? This and other vector methods were excluded due to limited usefulness.

arif-js:
> Hi @arif-js, can you share the use case for this? This and other vector methods were excluded due to limited usefulness.

Hey @ankane! Understood. I noticed it was missing for sparsevec. Though I haven't verified it in practice, the only case I can think of is checking dims on an untyped sparsevec column too large to cast to vector, and a typmod likely covers that. Happy to close if you'd rather keep the surface as is.

ankane:
Thanks @arif-js, I think it's better to hold off until there's a stronger use case.

